PR #286, opened 2024-10-15
Fix PHP Warning  - undefined `$wp_version`

jameshobden:
`$wp_version` would always be undefined as it wasn't loaded in local scope. Updated reference to global variable to fix.


brtkwr (approved):
LGTM

PR #287, opened 2024-10-15
DOMNodeInserted is deprecated, use MutationObserver instead

jameshobden:
DOMNodeInserted and will be dropped in future, use Mutation Observer instead. 

https://developer.mozilla.org/en-US/docs/Web/API/MutationEvent


brtkwr (approved):
LGTM thank you so much, really appreciate these!

PR #288, opened 2024-10-15
Comparison operator should be `OR`, not `AND`

jameshobden:
Comparison operator should be `OR`, not `AND`. Otherwise we try to run logic if just one of the parameters is true, resulting in PHP warning on unset `$_POST` value


brtkwr (approved):

brtkwr:
LGTM thank you!

PR #289, opened 2024-10-17
Fix PHP warning undefined $body

jameshobden:
This line is redundant as it is trying to access `$body` which is not yet defined, triggering a PHP Warning. Later in this method we define body (line 2314) from the JSON response and correctly reference it anyway.


jameshobden:
Sorry for all the PRs!

brtkwr:
> Sorry for all the PRs!

Not at all, really appreciate these! I am trying to figure out how to make the CI work with external PRs as they don't seem to play well with self-hosted runners.

brtkwr:
I've bundled these fixes into a new release: https://github.com/two-inc/woocommerce-plugin/releases/tag/2.19.3

Again, really appreciate them!

PR #293, opened 2024-11-14
CET-261/fix: Remove redundant 1-click WooCom set up code from the plug in

brtkwr:
**Context**

We used to have big dreams of 1-click WooCommerce. They were misplaced dreams, and knowing what we know now, a 1 click set up isn't possible.

**Requirements:**

Remove the old code


SalmanTwo (approved):
LGTM

PR #296, opened 2024-11-28
CET-528/feat: Refund order with Two when order status is updated to Refunded

brtkwr:
Also ensure compatibility with Woocommerce 8 and 9 which now use HPOS.


SalmanTwo (approved):
LGTM

PR #298, opened 2025-02-12
Handle checkout with no product post object

jameshobden:
It's possible for a user to checkout with custom line items in the basket that don't have a related product (post). This update handles this edge scenario by first checking if the product ID exists and falling back to standard values


class/WC_Twoinc_Helper.php
```diff
@@ -265,11 +265,26 @@ public static function get_line_items($line_items, $shippings, $fees, $order)
 
                 $tax_rate = WC_Twoinc_Helper::get_item_tax_rate($line_item, $order);
 
-                $image_url = get_the_post_thumbnail_url($product_simple->get_id());
+                // Check if product exists and is a valid object. If not, use fallback values.
+                if ( ! is_object($product_simple) ) {
+                    $name = method_exists($line_item, 'get_name') ? $line_item->get_name() : 'Item';
```
brtkwr, line 270: Some of these values are optional in the schema. It might be better to omit them altogether. Ref: https://docs.two.inc/openapi/checkout-api/#operation/create_order_handler.post!path=line_items&t=request

jameshobden: The simplest way I found of tweaking it was to set the variables either '' or defined, otherwise we end up with a load of ternary operators checking is_object($product_simple) inside the array definition which gets a bit messy. 

Of course if you know a better way please do make changes! Most important thing for us is just getting it to support it one way or another.

brtkwr (approved):
Sorry for the slow review! This slipped off my radar! LGTM

brtkwr:
Superseded by #307

PR #307, opened 2025-06-30
Catch edge case where the line item isn't an object

jameshobden:


brtkwr (approved):

jameshobden:
I think because I'm not a member of the two-inc org the checks fail

PR #323, opened 2026-06-11
TWO-24751/feat: Payment terms chip selector and offset pricing fee

dgjlindsay:
## What

Buyer-selectable invoice terms at checkout (chip selector, Magento parity) + optional offset pricing: the merchant's service fee passed to the buyer as a separate line, in three variants (simple 100%, partial percentage, incremental vs a reference term).

## Design

- **No plugin-side fee arithmetic.** Offset settings post to `POST /v1/pricing/order/fee` as a `buyer_fee_share` block; the backend's answer is charged verbatim (mirrors Magento's SurchargeCalculator pattern).
- **Business logic in PHP, JS renders only** (TWO-24751 checkout-type separation requirement; Gutenberg port needs only a new presentation layer). New `WC_Twoinc_Payment_Terms` class owns term resolution, selection validation, fee quoting, the cart fee, and payload terms. `twoinc.js` renders chips from two wc-ajax endpoints (`two_term_fees`, `two_select_term`).
- **Term availability behind a single seam.** `get_available_terms()` = brand `available_terms` (new key, 14/30/60/90 for Two) ∩ merchant admin subset. The planned backend term-availability surface replaces only this method — see the design-constraint comment on TWO-24751. **Deliberate divergence from the ticket text:** the admin day list is constrained to the brand set rather than free-form, so the future backend swap can't widen what a brand allows.
- **Fee enters the cart as a native WC fee** (taxed by store tax handling → order-internal net+tax=gross stays consistent; same posture as Magento's store-configured surcharge tax rate). It flows into the Two payload via the existing fee line-item path. Cart changes re-trigger quote + render on `updated_checkout`.
- **Fail-soft quotes**: a failed fee fetch renders chips without labels; checkout never blocks. Single term renders one non-interactive chip.
- Selected term + offered set ride the create-order payload as `terms` / `available_terms`; selection also posts with the checkout form (hidden field) so order-pay-page flows work sessionless.

## Tests

6 new unit specs (PHP 8.2 via `make test-unit`, all green): brand∩admin resolution, default fallback, the three buyer_fee_share shapes + malformed-percentage fallback, payload terms presence/absence, invalid-POST fallback.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -178,3 +178,36 @@
   position: absolute;
   right: 4px;
 }
+
+/* Payment terms chip selector (TWO-24751) */
+.twoinc-term-chips {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 8px;
+  margin: 8px 0;
+}
+.twoinc-term-chips.hidden {
+  display: none;
+}
+.twoinc-term-chip {
+  display: inline-flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 6px 14px;
+  border: 1px solid #c4c9e8;
+  border-radius: 16px;
+  background: #fff;
+  cursor: pointer;
+  line-height: 1.3;
+}
+.twoinc-term-chip--selected {
+  border-color: #3043d1;
+  background: #eef0fb;
+}
+.twoinc-term-chip--single {
+  cursor: default;
+}
+.twoinc-term-chip__fee {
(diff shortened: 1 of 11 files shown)


dgjlindsay:
Closing: superseded. Commits fully contained in `staging` (integrated via #329–#339). Empty delta vs staging. Reopen if wrong. — closed by Claude on behalf of Doug

PR #324, opened 2026-06-11
INF-1598/ci: repoint actions to two-inc/actions monorepo

brtkwr:
Repoints GitHub Action references from the legacy `two-inc/<x>-action` repos to the consolidated [`two-inc/actions`](https://github.com/two-inc/actions) monorepo (INF-1598). No behaviour change - same actions, new path.

https://linear.app/tillit/issue/INF-1598

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -16,6 +16,6 @@ jobs:
       contents: read
       pull-requests: write
     steps:
-      - uses: two-inc/pre-commit-action@main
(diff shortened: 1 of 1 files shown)


brtkwr:
closing - public repos cannot consume the private two-inc/actions monorepo (github blocks public->private action resolution even with org-access). woocommerce-plugin stays on two-inc/pre-commit-action which remains public + archived and keeps working

PR #325, opened 2026-06-12
INF-1598/ci: repoint pre-commit to two-inc/actions-public monorepo

brtkwr:
Repoints `two-inc/pre-commit-action` -> `two-inc/actions-public/pre-commit` (the new public companion to the private `two-inc/actions` monorepo).

Public repos cannot consume the private `two-inc/actions` monorepo (GitHub blocks public->private action resolution even with org-access), so the two public-safe actions (pre-commit, claude-reviewer) live in the public `two-inc/actions-public`.

Part of INF-1598 actions consolidation.

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -16,6 +16,6 @@ jobs:
       contents: read
       pull-requests: write
     steps:
-      - uses: two-inc/pre-commit-action@main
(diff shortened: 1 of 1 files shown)


Puvendhan (approved):
Approved: INF-1598 mechanical repoint to consolidated actions monorepos.

PR #342, opened 2026-07-06
TWO-24812/feat: source offerable payment terms from GET /v1/merchant

dgjlindsay:
## What

Swaps the term-availability source from the hardcoded brand-file list to `available_terms` on `GET /v1/merchant/{id}` (the backend resolves it from the merchant's pricing packages). Backend owns which terms exist; the admin narrows; the checkout seam (`WC_Twoinc_Payment_Terms::get_available_terms`, the single term-list read by design) intersects the saved admin subset with the backend list at read time — a term the backend withdraws drops out even while a stale admin selection still ticks it. Custom term stays unioned in, unchanged.

## How the record resolves

New `WC_Twoinc::get_merchant_available_terms()`, following the `get_platform_minimum_order()` option-cache pattern (15 min TTL, 10s request cap on the render path), with one deliberate difference: the stored list is **only overwritten by a successful response carrying an `available_terms` array** — an API blip (or an older backend omitting the field) serves the last-known list for another TTL instead of blanking checkout terms. An explicit `[]` from the backend is honoured (nothing offerable). Unresolved (fresh install / no API key) → no presets; admin screen explains terms load once a valid API key is configured; checkout falls through to the account default term. Same degrade posture as Magento's `SettingsProvider`/`RecordProvider`.

## Behavioural note for existing shops

The staging backend currently returns `available_terms: [0, 7, 15, 20, 30, 60, 90]` (verified live; 0 is filtered by the `>0` guard, matching Magento). **14 is not in the backend list**, so a shop whose saved selection ticks the old brand-default 14-day term stops offering it once this deploys — that's the intended semantics (backend is the source), but worth knowing when the staging shop's chips change.

The brand file no longer carries `available_terms`; an overlay defining it has no effect (the partner edition needs no term override — the backend seeds partner defaults per merchant).

## Verified

- Unit suite: 49/49 pass locally (PHP 8.5), including two new tests — backend∩admin resolution (withdrawn term, unresolved set, seam guard for gateway fakes) and the fetch/normalise/cache/serve-stale protocol.
- `php -l` clean on all changed files; prettier pre-commit clean.
- Live staging probe: `GET /v1/merchant/{id}` returns `available_terms` for the e2e merchant (backend consumer contract confirmed live).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ coverage/
 # Playwright
 tests/e2e/test-results/
 tests/e2e/playwright-report/
+.review/
--- a/brands/two.php
+++ b/brands/two.php
@@ -41,11 +41,10 @@ return [
     // (merchant-saved titles always win). sprintf'd with the invoice
     // day count, so a brand default may carry one %s.
     'title_default' => 'Business invoice - %s days',
-    // Terms the brand may offer in the checkout chip selector (the
-    // merchant narrows the set in settings; WC_Twoinc_Payment_Terms is
-    // the only reader). Mirrors the Magento brand descriptor's
-    // available_payment_terms.
-    'available_terms' => [14, 30, 60, 90],
+    // NOTE: the offerable payment-term list no longer lives in the brand
+    // file — it is sourced per merchant from `available_terms` on
+    // GET /v1/merchant (TWO-24812). A brand overlay defining
+    // 'available_terms' has no effect.
     // Increments the buyer surcharge line may be rounded to, offered in
     // the admin Rounding Step dropdown (the merchant picks one; the None
     // basis disables rounding). WC_Twoinc::get_rounding_step_options is
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -316,29 +316,42 @@ if (!class_exists('WC_Twoinc')) {
          * `available_terms` on GET /v1/merchant/{id} — the backend resolves
          * them from the merchant's pricing packages, so this is the
          * authoritative set the admin narrows from (TWO-24812; the brand
-         * file no longer carries a term list). Empty means the set cannot
-         * currently be resolved (no API key yet, or no successful fetch) —
-         * no terms are offered and the backend applies the account default,
-         * the same degrade posture as Magento's SettingsProvider.
+         * file no longer carries a term list). Empty means either the set
+         * cannot currently be resolved (no API key yet, no successful fetch
+         * yet) or the backend explicitly returned an empty list (nothing
+         * offerable) — in both cases no terms are offered and the backend
+         * applies the account default, the same degrade posture as
+         * Magento's SettingsProvider.
          *
-         * Cached in options for 15 minutes following the
-         * get_platform_minimum_order() pattern, with one difference: the
-         * stored list is only overwritten by a successful response carrying
-         * an `available_terms` array. A fetch failure (or an older backend
+         * By default this only READS the cached option — it never blocks on
+         * HTTP, because the terms seam is reached from contexts that must
+         * not stall (the gateway constructor via init_form_fields, cart
+         * totals, wc-ajax). A refresh (15-minute TTL, 10s request cap) runs
+         * only where `$refresh = true` is passed: the checkout render
+         * bootstrap and the admin payment-terms field render. The stored
+         * list is only overwritten by a successful response carrying an
+         * `available_terms` array; a fetch failure (or an older backend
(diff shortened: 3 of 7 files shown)


dgjlindsay:
## Self-review: adversarial review summary (per the self-review lane for plugin repos)

**Slate:** 3 personas, non-overlapping mandates — Leia (Staff Engineer / long-term maintainability), Han (SRE / failure modes, races, cache lifecycle), Vader (adversarial QA / user breakage + test fidelity). Yoda dropped (thin framework surface), Obi-Wan dropped (no new security surface — API-key handling untouched; password/secret handling not in scope).

**Convergence: 4 rounds, 11 → 6 → 3 → 0 substantive findings.**

### Round 1 (11 substantive) — all fixed in `8b5964b`
- Fetch escaped the checkout render path into the gateway constructor (init_form_fields) and cart totals → **refresh-gating**: accessor is cache-only by default, TTL fetch only at the checkout render bootstrap + admin field render; apply_cart_fee gates reordered (Han, Vader convergent).
- No cache invalidation on API-key/merchant-id change → serve-stale could pin the OLD merchant's terms indefinitely → **identity-change invalidation** in process_admin_options + verify_api_key (Han).
- Fresh-install first save threw the mandatory-terms error; save-time filtering silently erased ticks/grid rows when the backend list shifted → **non-destructive validators** (unresolved list keeps stored; narrowed list unions stored; grid preserves unrendered rows) (Han, Vader convergent).
- Silent mid-checkout re-price when the selected term is withdrawn → **process_payment abort with buyer-facing notice** (Vader).
- intval on malformed backend elements minted a phantom "1 day" term → is_numeric guard (Han, Vader convergent). Plus docblock/copy corrections (Leia).

### Round 2 (6 substantive) — all fixed in `e1a9d86`
- The checkout-render refresh wrote the **whole WC settings blob** from a frontend request every TTL — a stale in-memory snapshot could silently revert a concurrent admin save wholesale → cache moved to **two dedicated brand-prefixed wp_options** (Han, Vader convergent).
- Grid preservation used the live term set as a proxy for "rendered", which erases rows in the widening direction — including a **deterministic same-save re-tick case** (sibling field validates first) → preservation now keys on the **POSTed row keys**; null POST preserves everything (Han, Vader convergent, repro'd).
- Default-term validator lacked the sibling's unresolved-list guard → blanked the stored default (Leia).
- Single-term checkouts had a dead zone in the re-price guard (chips JS never renders, nothing posts) → **server-side hidden term input**, with the chips JS input later in the DOM winning the POST when chips render (Vader).
- is_selector_visible restored as the single encoding of the >1-terms rule (Leia).

### Round 3 (3 substantive) — all fixed in `523f5a0`
- "Clear options on deactivation" left the two new dedicated options orphaned → invalidation added to on_deactivate_plugin (Leia).
- A saved **custom term punched through the default-term degrade path** (posted custom made $offered non-empty → stored default silently repointed on an API-blip save) → keep-stored guard moved before the custom-append; repro'd then fixed (Vader).
- The no-API-key test assertion was **dead — proven by mutation** (deleting the guard kept the suite green): TTL now expired + sentinel response queued so the guard is genuinely load-bearing (Vader).

### Round 4 — CLEAN ROUND ✅
- **Leia: zero substantive.** All round-3 fixes verified in place; suite green.
- **Vader: zero substantive.** Every round-3 fix re-verified by mutation (guard deletion → suite fails at the exact designed assertion; round-2 bypass shape reintroduced → fails on the exact repro). Working tree confirmed restored.
- **Han: zero substantive in the diff.** One pre-existing find outside this PR: the deactivation hook has been registered on the wrong basename forever, so the whole clear-options feature (round-3 fix included) is currently unreachable — ticketed as TWO-25028 rather than grown into this PR, per Han's own recommendation.

### Disputed / deferred (with reasoning, not silently dropped)
- Triple independent GET /v1/merchant fetch + sibling caches still in the settings blob + TTL stampede → **TWO-25024** (consolidation; deliberately out of scope for diff size).
- Term `0` semantics in the backend list and prepopulate-shortest now landing on 7 days → product questions noted on TWO-25024.
- Key-rotation + tick-edit in one save discards the tick edits silently (defensible: ticks referenced the old merchant's list; form re-renders correctly) — accepted residual.

**Gate met:** one full round with zero substantive findings on the diff; unit suite 55 tests green locally (PHP 8.5) and in CI (7.4 + 8.2), e2e + style green. Merging under the self-review lane (repo family designated self-reviewable 2026-07-06).

*— adversarial self-review conducted and summarised by Claude on Doug's behalf*

PR #344, opened 2026-07-07
TWO-25028/fix: register deactivation hook from the main plugin file

dgjlindsay:
## Problem

The deactivation cleanup hook was registered inside `class/WC_Twoinc.php` as

```php
add_action('deactivate_' . plugin_basename(__FILE__), [$this, 'on_deactivate_plugin']);
```

`__FILE__` there is the class file, so the hook name resolves to `deactivate_tillit-payment-gateway/class/WC_Twoinc.php` — but WordPress fires `deactivate_tillit-payment-gateway/tillit-payment-gateway.php` (the main plugin basename). The hook never fired, so the **Clear options on deactivation** setting has been dead since it shipped.

## Fix

- `register_deactivation_hook(__FILE__, 'twoinc_on_deactivate_plugin')` in the main plugin file, delegating to `WC_Twoinc::get_instance()->on_deactivate_plugin()` — which already clears the settings blob **and** the dedicated term-cache options (`{prefix}_merchant_available_terms{,_checked_on}`, TWO-24812).
- Removed the dead registration from the gateway constructor, with a comment pointing at the main-file registration so it doesn't creep back.
- Unit test `testDeactivationCleanupClearsSettingsAndTermCache` covers both toggle states (off = everything kept; on = blob + both cache options deleted).

## Known limitation (pre-existing)

If WooCommerce is deactivated first, the plugin file returns before registering the hook, so deactivating Two afterwards skips cleanup. Same guard gates the whole plugin; not new behaviour.

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -105,8 +105,9 @@ if (!class_exists('WC_Twoinc')) {
                 // Verify API key quietly
                 add_action('admin_enqueue_scripts', [$this, 'verify_api_key_action']);
 
-                // On plugin deactivated
-                add_action('deactivate_' . plugin_basename(__FILE__), [$this, 'on_deactivate_plugin']);
+                // Deactivation cleanup is registered in the main plugin file
+                // (register_deactivation_hook) — a registration here would key
+                // the hook off THIS file's basename and never fire (TWO-25028).
 
                 // Add js css to admin page
                 add_action('admin_enqueue_scripts', [$this, 'twoinc_admin_styles_scripts']);
--- a/tests/unit/run.php
+++ b/tests/unit/run.php
@@ -1008,6 +1008,18 @@ final class BrandConfigSpec
             $GLOBALS['__twoinc_test_options'][$checked_option] = 999;
         };
 
+        // Toggle key absent from the settings blob: the state every merchant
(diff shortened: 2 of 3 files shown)


dgjlindsay:

**Slate:** Han (SRE/failure modes), Vader (edge cases/data loss) — 2 lenses, small fix PR.

**Round 1:** no BLOCKER/BUG. Vader verified the core claim (the old `add_action('deactivate_' . plugin_basename(__FILE__))` resolved to the class file's basename and never fired; the fix registers from the main file, matching WP's hook name) and that the cleanup deletes exactly the prod option keys. Han flagged release-hygiene observations, all accepted as notes rather than code changes:
- Network-activated multisite: cleanup runs in network context and clears the main site only — pre-existing shape, noted on the ticket.
- A dead-since-ship destructive toggle goes live with this fix: merchants who once ticked "clear settings on deactivation" will now actually lose settings (incl. API key) on their next manual deactivate. Default is `no`; plugin *updates* don't fire the hook. Needs a release-note callout when this ships.

**Fixes applied (3e82d2f):** test now also pins the toggle-key-absent case (no wipe) — the state every merchant who never touched the toggle is in.

**Round 2:** clean. CI green.

PR #345, opened 2026-07-07
TWO-25024/refactor: one request-memoised GET /v1/merchant fetch for all consumers

dgjlindsay:
## Problem

Three independent fetch-and-cache readers of the same `GET /v1/merchant/{id}` payload, with divergent TTLs/timeouts/failure postures:

| Reader | TTL | Timeout | On failure | Cache |
|---|---|---|---|---|
| `get_merchant_default_days_on_invoice` | 1h | 30s (uncapped) | default 14 | settings blob |
| `get_platform_minimum_order` | 15m | 30s (uncapped) | null | settings blob |
| `get_merchant_available_terms` (TWO-24812) | 15m | 10s | serve stale | dedicated wp_options |

Worst-case cold render against a hanging API stacked ~70s of serial timeouts; healthy API got three identical GETs per TTL alignment. The blob writes were the sharper hazard: a frontend TTL-expiry write via `WC_Settings_API::update_option` rewrites the **whole** settings array from this request's in-memory snapshot, silently reverting a concurrent admin settings save.

## Change

- **`fetch_merchant_record()`** — one wire fetch per PHP request (memoised, failures included), 10s cap, shared by all three consumers. Max one capped stall per request.
- **Consumers keep their own TTL + degrade posture** (1h/15m/15m; days defaults to 14 and serves stale, minimum blanks to null, terms serve stale) — only the wire is shared.
- **Blob migration**: `days_on_invoice` and `platform_minimum_order` caches move to dedicated brand-prefixed wp_options (same pattern as the term cache). Cold-start on upgrade: first TTL check refetches; old blob keys are orphaned and go with the blob on cleanup.
- **Stampede guard**: `checked_on` bumps *before* the fetch — concurrent requests at expiry serve stale instead of piling on.
- **`invalidate_merchant_record_caches()`** (was `invalidate_merchant_available_terms`) drops all six options + the memo; used by identity-change invalidation and deactivation cleanup.

## Tests

New `testMerchantRecordFetchSharedAcrossConsumersAndOffTheBlob`: three consumers, one `make_request`; caches land in dedicated wp_options and never the blob; API-down request costs exactly one more call with each consumer's degrade posture intact. Existing term-cache tests updated only where they simulate request boundaries (`$expire` also resets the per-request memo — TTL expiry can only happen across requests).

## Not addressed here (tracked on the ticket)

Product questions from the TWO-24812 review: whether backend term `0` ("due on receipt"?) should be offerable, and whether prepopulate-shortest ticking 7 days for never-configured merchants is acceptable. Behaviour unchanged.

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -17,6 +17,11 @@ if (!class_exists('WC_Twoinc')) {
     {
         private static $instance;
 
+        // Per-request memo for GET /v1/merchant/{id} (TWO-25024): one wire
+        // fetch per request shared by every consumer, failures included.
+        private static $merchant_record = null;
+        private static $merchant_record_fetched = false;
+
         // BC-frozen: external integrations may read these constants, but all
         // runtime reads go through WC_Twoinc_Brand so overlays can rebrand.
         // They mirror brands/two.php; tests/unit pins them against drift.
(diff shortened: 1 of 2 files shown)


dgjlindsay:

**Slate:** Leia (maintainability), Han (production failure modes), Vader (data-shape/boundaries).

**Round 1:** no BLOCKER/BUG. Convergent (Leia+Han): the three consumers ran the configured-check/TTL-bump dance in three different orders — days_on_invoice bumped its clock before checking configuration, so unconfigured installs wrote an option hourly. Vader confirmed the blob-clobber fix is genuine (no residual readers/writers of the old blob keys), memo statics are per-request under PHP-FPM, and each consumer's degrade posture survived the refactor unchanged.

**Fixes applied (6afe0c6):** guard order aligned across all three consumers (configured → TTL → bump → fetch); `$response['body']` hardened on the shared fetch path; memo-reset and invalidation docblocks corrected (production reaches the reset only via `invalidate_merchant_record_caches()`; new-consumer registration instruction added).

**Declined, with reasons:**
- Minimum's no-serve-stale posture (failure erases the cached minimum): pre-existing, documented behaviour — changing it inside a refactor PR would smuggle a semantic change. Candidate follow-up.
- TTL magic numbers → class const: deferred; noted for a fast-follow now a 4th copy exists in #346.
- Known deltas worth naming: one shared fetch means all consumers degrade *together* on a flaky API (the point of the ticket, but a real behavioural change on intermittent outages); the read-then-write stampede guard is best-effort, not atomic; the constructor still does a synchronous once-per-TTL wire fetch (10s cap) — async refresh is follow-up material.

**Round 2:** clean (verified alongside #346's rebase). CI green.

PR #346, opened 2026-07-07
TWO-24954/feat: read + enforce funding-partner surcharge cap from GET /v1/merchant

dgjlindsay:
**Stacked on #345** (TWO-25024 merchant-record provider) — merge that first; this PR's diff vs its base is the surcharge-cap feature only.

## What

Feature B consumer (Woo base plugin had no surcharge-limit concept; built whole, mirroring Magento):

1. **Read**: `get_merchant_surcharge_limit()` — `surcharge_limit_amount`/`_currency` (TWO-24950, merged) read off the shared request-memoised merchant-record fetch from #345, so zero extra round-trips. Cached 15m in dedicated brand-prefixed wp_options; **serve-stale on failure** (a stale cap still enforces — dropping it on an API blip would let an over-limit fee through). Refresh gated to admin contexts: grid render + grid save validation.
2. **Enforce**: `validate_two_surcharge_grid_field` rejects per-term `fixed` above the cap with a message naming it ("…exceeds the maximum of EUR 25."), mirroring Magento's `SurchargeGrid::validateValue`. Currency guard: enforced only when the cap currency matches the store currency — Woo does no FX conversion (Magento converts to base); on mismatch the backend enforces alone.
3. **Help text**: conditional sentence under the grid using Magento's `getFixedLimitLabel` wording ("Enter the amount you want to charge your customer. Max EUR 25.") — omitted entirely when no cap is configured, so no-limit brands see no change.

## Tests

`testSurchargeGridEnforcesMerchantFixedCap`: at-cap allowed, above-cap rejected naming the cap, percentage/per-order-limit columns unaffected, currency-mismatch skip, no-cap unchanged. The shared-provider test now covers the fourth consumer (still exactly one wire fetch; serve-stale on API-down).

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -365,6 +365,70 @@ if (!class_exists('WC_Twoinc')) {
         }
 
         /**
+         * The merchant's fixed-fee surcharge cap from GET /v1/merchant
+         * (surcharge_limit_amount/_currency — the funding partner's upper
+         * bound on what a merchant may pass on per order, TWO-24950), as
+         * ['amount', 'currency'] or null when none is configured. Mirrors
+         * Magento's SettingsProvider::getSurchargeLimit (TWO-24954).
+         *
+         * Cached for 15 minutes in dedicated brand-prefixed wp_options with
+         * the term-cache posture: read-only by default (the value gates an
+         * admin save, never a checkout render), refresh only where
+         * $refresh = true is passed — the surcharge grid render and its
+         * save-time validation. Serve-stale on fetch failure: a stale cap
+         * still enforces; dropping it on an API blip would let an
+         * over-limit fixed fee through.
+         *
+         * @param bool $refresh allow a TTL-gated fetch on this call
+         * @return array|null
+         */
+        public function get_merchant_surcharge_limit(bool $refresh = false)
+        {
+            $limit_option = WC_Twoinc_Brand::prefixed_name('merchant_surcharge_limit');
+            $checked_option = WC_Twoinc_Brand::prefixed_name('merchant_surcharge_limit_checked_on');
+            $checked_on = get_option($checked_option);
+
+            if ($refresh && (!$checked_on || ((int) $checked_on + 900) <= time())) {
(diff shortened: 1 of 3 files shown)


dgjlindsay:

**Slate:** Leia (maintainability/Magento parity), Han (production failure modes), Vader (boundary prosecution).

**Round 1** — one convergent finding (3/3 reviewers): the grid help text claimed "Max EUR 25" even in stores whose currency doesn't match the cap — exactly where save-validation is deliberately skipped (Woo does no FX conversion; Magento converts or warns). Also: displayed 2dp max could be unenterable against full-float enforcement (cap 25.555 → "Max 25.56" but 25.56 rejected); duplicated label-formatting expression; Han showed the grid preservation loop lets an unrendered (withdrawn-term) row bypass the cap if the term later returns.

**Fixes applied (3634f4d):** help text now carries the same currency guard as validation — the grid never claims a limit it won't enforce; cap rounded to 2dp at read/write so label == enforcement; `format_surcharge_limit_label()` single source for help text + error; preservation-loop bypass documented as deliberate (admin can't fix an invisible row; save-time check is Magento UX parity; the funding partner's backend is the authoritative enforcement point). New tests: comma-locale over-cap input, zero-amount cap record = no limit, currency-mismatch omits the Max sentence.

**Declined/accepted as-is:**
- Save-time-only enforcement (pre-existing over-cap values keep charging until re-save): byte-for-byte Magento parity; backend-side enforcement is the backstop and its coverage is tracked separately (TWO-25006/25007 family).
- Cap binds the fixed column only (percentage/limit uncapped): exact Magento parity; product question, not a code fix.
- Stale-lower-cap can reject a legit save for ≤15 min after ops raises the cap (grid render stamps `checked_on`, save-time refresh is TTL-gated): self-healing, error names the cap it enforced.
- Magento int-casts the cap, Woo enforces the (2dp-rounded) float — Woo is the more faithful reading; divergence now visible in code comments.

**Round 2:** clean — guard identity, single-source label, 2dp idempotence, sibling guard-order alignment and test genuineness all verified. CI green.

PR #347, opened 2026-07-07
TWO-25015/ci: merchant-upgrade smoke test (latest release → HEAD)

dgjlindsay:
## What

New `upgrade-smoke` workflow proving the merchant upgrade lands cleanly on every PR:

1. Boot WordPress 6.8 + WooCommerce 9.9.5 (same pins as the e2e harness in `docker-compose.yaml`).
2. Install the **latest released tag** (resolved via `git ls-remote`), activate, seed gateway settings.
3. Replace the plugin files with this branch's HEAD — the exact file swap a WordPress plugin upgrade performs on a live shop.
4. Verify: plugin still active, `WC_Twoinc` loads and `get_instance()` boots, pre-upgrade settings survived, storefront renders without a fatal.

## Conventions ported from magento-plugin's upgrade-smoke (TWO-24998)

- **`git ls-remote` over the GitHub tags API** — no token, no per_page cap, and a transport error fails the step instead of skip-greening on empty JSON.
- Fail-loud asserts, no readiness sleeps (docker health checks + `until` loops).

## Deviations from the ticket, reasoned

- **FROM = latest release, not "oldest-supported"**: Woo has no defined support floor, and old tags won't necessarily install on current WP/WC. The latest published release is the version every live merchant actually upgrades from, and it moves forward automatically as releases ship. Magento used previous-major because a real 1.x→2.x boundary existed; Woo's line is single-major.
- **Checksum, not version comparison, proves the swap**: the HEAD version header legitimately equals the latest tag between releases, so `wp plugin get --field=version` can't distinguish prior from HEAD; an md5 of the main file can.
- **No schema-migration leg**: WordPress has no `setup:upgrade` equivalent — the upgrade surface is file replacement + option-data compatibility, which is exactly what's exercised.

### DIFF
--- a/.github/workflows/upgrade-smoke.yaml
+++ b/.github/workflows/upgrade-smoke.yaml
@@ -1,0 +1,152 @@
+name: Upgrade smoke
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: upgrade-smoke-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  # Merchant-upgrade smoke (TWO-25015): install the latest released tag the
+  # way a merchant runs it, then replace the plugin files with this branch's
+  # HEAD — the exact file swap a WordPress plugin upgrade performs — and
+  # verify the plugin stays active, loads, and keeps its settings. WordPress
+  # has no schema-migration step (unlike Magento's setup:upgrade), so the
+  # upgrade surface is: file replacement + option-data compatibility.
+  upgrade-smoke:
+    name: Upgrade smoke (latest release → HEAD)
+    runs-on: ${{ vars.RUNNER_STANDARD }}
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Resolve latest release tag
+        id: prior
+        run: |
+          set -o pipefail
+          # git ls-remote (not the GitHub tags API): no token, no per_page
+          # cap, and a transport error FAILS the step rather than yielding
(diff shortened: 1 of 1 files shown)


dgjlindsay:

**Slate:** Han (CI flakiness/false-green), Vader (does-it-prove-what-it-claims).

**Round 1** — the storefront probe was vacuous: the site installed as `http://wp` while the probe curled `http://localhost/`, so WP answered with a canonical 301 that `curl -sf` (no `-L`) counts as success without rendering a page. Also: fixed docker resource names could collide across *concurrent PRs* (the concurrency group is per-ref) and one job's teardown could rm another's live containers; both `until` waits were unbounded; the "settings survived" assertion proves less than its framing (a file swap can't touch wp_options); the swap-fidelity claim overstated (HEAD leg = git-tracked tree incl. dev files, release leg = export-ignore'd archive, merchants get the `.distignore` zip); checkout path untested post-upgrade.

**Fixes applied (86813b7):** install URL now `http://localhost` so the render probes exercise the template for real; post-upgrade checkout-page render check added (explicit page-id assert, no xargs); all docker resources namespaced by run id; waits bounded at 60 iterations with `::error` + container logs; settings and fidelity comments now claim only what they prove.

**Round 2 (fbd38bd):** verifier confirmed every fix line-by-line (incl. the non-numeric page-id guard behaviour) and found one parity gap — the two network downloads (WooCommerce install, release tarball) lacked the retry hardening #348 got. Both now have bounded retries.

**Accepted as-is:** the checksum degenerates when the latest tag's main file is byte-identical to HEAD (true today, 2.23.9) — it proves deployed==HEAD content, not that the swap executed; acceptable for a smoke. Pins (`mariadb:10.6`, WC 9.9.5) drift in lockstep with docker-compose.yaml.

CI green.

PR #348, opened 2026-07-07
TWO-25016/ci: version-combination support matrix (light two-axis scope)

dgjlindsay:
## Scope decision (AC 1 — decided before building, documented in the workflow header)

**Light two-axis subset, not Magento's full N×M.** Magento's matrix multiplies Magento-minor × PHP because Adobe's support policy + the CI images make each combination a genuinely different runtime. For Woo the equivalent multiplication (WP × WC × PHP) mostly re-tests WooCommerce itself — WooCommerce gates its own WP/PHP support — so each axis is tested where it's cheapest and most meaningful:

- **PHP axis** → `unit-tests.yaml` matrix extended from `[7.4, 8.2]` to **7.4–8.4** (declared floor through every currently-supported minor). The tiny dependency-free suite makes full span effectively free, and this is where PHP-syntax/behaviour breakage actually surfaces.
- **WooCommerce axis** → new `support-matrix.yaml`: boot-smoke of plugin HEAD against the **latest WC release** and the **previous minor line**, both discovered live from the wordpress.org plugin API (no hand-maintained pins). Each leg: activate, `WC_Twoinc` boots, gateway present in WooCommerce's registry, storefront + checkout pages render without fatals.

## Honest-CI conventions carried over from Magento's matrix (TWO-24998)

- Discovery failure (API/transport) **fails** the job — never an empty matrix that skip-greens having tested nothing.
- Every emitted leg genuinely runs; `fail-fast: false` so one leg's failure doesn't mask the other's result.

## Not covered, deliberately

- No WC floor below previous-minor: the plugin declares no minimum WC version anywhere (readme says "tested up to", compose pins 9.9.5). If we want a support floor tested, that's a product declaration first — happy to add a third leg once one exists.
- No WP-version axis: WP ≥6 declared, but plugin surface is WC-API-bound; the WC axis exercises the integration seam.

### DIFF
--- a/.github/workflows/support-matrix.yaml
+++ b/.github/workflows/support-matrix.yaml
@@ -40,7 +40,8 @@ jobs:
         id: wc
         run: |
           set -o pipefail
-          if ! info=$(curl -fsSL "https://api.wordpress.org/plugins/info/1.0/woocommerce.json"); then
+          if ! info=$(curl -fsSL --retry 5 --retry-all-errors --retry-delay 3 \
+            "https://api.wordpress.org/plugins/info/1.0/woocommerce.json"); then
             echo "::error::support-matrix: could not query wordpress.org plugin API"
             exit 1
           fi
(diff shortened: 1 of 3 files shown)


dgjlindsay:

**Slate:** Han (CI flakiness/drift), Vader (does-each-leg-prove-support).

**Round 1** — three real false-green/flake holes: (1) same canonical-301 vacuous render probe as #347 (site `http://wp`, probe `localhost`); (2) the checkout leg piped the page id through `xargs`, so an empty/0 id ran the curl **zero times** and passed; (3) single un-retried curl to the wp.org API on every push — a transient 429/5xx would red unrelated PRs. Plus: parallel matrix legs shared fixed docker names (collision on a shared runner); unbounded waits; readme.txt declared `Requires PHP: 7` while the matrix floor (and real code floor — strict_types, typed signatures) is 7.4.

**Fixes applied (e9ffb9e):** install URL → `http://localhost`; explicit checkout-page-id assert (fails loud on empty/0); `--retry 5 --retry-all-errors` on discovery + bounded 3-attempt retry on plugin install; per-leg namespacing (`run_id-job-index`) with per-leg networks and `--network-alias` for inter-container DNS; bounded waits; `Requires PHP: 7.4`; unit-tests checkout action aligned to v6.

**Round 2:** clean — verifier walked every namespace reference, loop bound, retry failure path, and the garbage-page-id branch. CI green (both live-discovered WC legs).

**For the record (accepted scope costs):** the WC axis floats on both dimensions (unpinned WP image + live-discovered WC), so a green→red flip can be upstream drift rather than the PR's diff — if this ever becomes a *required* check and drift-blocking hurts, split a nightly canary from the per-PR run. PHP axis under real WP runs 8.2 only; 7.4–8.4 coverage is parse/load-level via the unit matrix (which does `require` the real class files). Documented in the workflow header.

PR #350, opened 2026-07-08
TWO-25035/feat: explicit environment mode + brand host template

dgjlindsay:
## Problem

The ABN staging shops (`woocom{,-dev}.staging.achterafbetalen.co`) show no merchant short name and never pick up payment terms: `is_twoinc_development()` only recognises `*.two.inc` dev hostnames, so those shops resolve the **Production** API host and every merchant-record fetch 401s with their staging key (verified live: same key 401s on the prod host, 200s on the staging host). The bump-then-fetch guard stamps `*_checked_on` anyway, masking the failure per TTL window.

Hardcoding brand staging domains in base would be a leak — base code must stay brand-agnostic.

## Fix — adopt the Magento pattern

Environment is an explicit mode, hosts come from a brand-owned template (mirrors `magento-plugin`'s `Config/Repository::getCheckoutApiUrl` + brand descriptor `checkoutUrlTemplate` + `Mode` source model):

- `checkout_env` = mode (`PROD`/`SANDBOX`/any custom value, e.g. `staging`), normalised by `WC_Twoinc_Helper::get_environment_mode()`.
- Host = `sprintf(brand's checkout_url_template, 'api'|'checkout' [+ '.mode'])` — Two brand template `https://%s.two.inc` reproduces today's prod/sandbox hosts exactly; overlays set their own domain in their brand file.
- Explicit non-default mode always wins. The dev-hostname sniffer stays as a **deprecated fallback** only while `checkout_env` is default — in-place wp.org upgrades and `*.staging.two.inc` shops keep today's behaviour untouched.
- Sole-trader signup URL rides the same template.
- `checkout_env` select preserves a stored custom mode as an option (Magento `Mode.php` trick) and is now always visible; the free-text test host shows only on sniffed dev environments.
- Hardcoded merchant-staging hostname list removed — all four hosts confirmed dead (3 have no DNS; staging.torn.no confirmed retired). `TWOINC_DEV_HOSTNAMES` remains the supported escape hatch.

## Rollout

1. Merge this + the abn overlay counterpart (template `https://%s.achterafbetalen.abnamro.nl`).
2. `wp option` `checkout_env=staging` on the ABN staging shops (and optionally the two.inc staging shops — sniffer covers them either way).

## Tests

66 unit tests green, incl. new: mode normalisation, template resolution (brand override), sniffer-vs-explicit-mode precedence matrix.

### DIFF
--- a/brands/two.php
+++ b/brands/two.php
@@ -20,6 +20,12 @@ return [
     'provider_full_name' => 'Two',
     'product_name' => 'Two',
     'sign_up_url' => 'https://portal.two.inc/auth/merchant/signup',
+    // sprintf template building environment hosts, mirroring the Magento
+    // brand descriptor's checkout_url_template: %s receives 'api' /
+    // 'checkout' in production, or 'api.<mode>' / 'checkout.<mode>' for
+    // any other checkout_env mode ('sandbox', 'staging', ...). Overlays
+    // set their own domain here — brand domains never live in base code.
+    'checkout_url_template' => 'https://%s.two.inc',
     'alert_email_address' => 'woocom-alerts@two.inc',
     'gateway_id' => 'woocommerce-gateway-tillit',
     'logo_url' => WC_TWOINC_PLUGIN_URL . 'assets/images/two-logo.svg',
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -186,9 +186,11 @@ if (!class_exists('WC_Twoinc')) {
             // TWOINC_DEV_HOSTNAMES env var, *.two.inc dev subdomains):
             // installs predating the explicit environment mode carry the
             // default mode and rely on the sniffed test host. Consulted only
-            // while checkout_env is still the default — an explicitly set
-            // mode always wins. Set checkout_env ('staging', 'sandbox', ...)
-            // instead of extending the sniffer.
+            // while the mode resolves to production — an explicit
+            // non-production mode ('staging', 'sandbox') always wins, while
+            // an explicit Production selection is indistinguishable from the
+            // never-configured default, so the sniffer still applies there.
+            // Set checkout_env instead of extending the sniffer.
             if (
                 WC_Twoinc_Helper::get_environment_mode($this) === 'production'
                 && WC_Twoinc_Helper::is_twoinc_development()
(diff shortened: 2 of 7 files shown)


dgjlindsay:

**Round 1** (3 personas): Leia/maintainability, Han/failure-modes+migration, Vader/adversarial-QA.

Findings and outcomes:
- **BLOCKER (Han): settings-API recursion.** `get_checkout_env_options()` called `$this->get_option()` from inside `init_form_fields()`; on any install whose settings blob lacks `checkout_env` (fresh installs, dev shops where the field was previously unset) `WC_Settings_API::get_option` re-enters `get_form_fields() → init_form_fields()` — infinite recursion. **Fixed** (69ee350): options builder reads the raw settings row; regression test uses a gateway whose `get_option` throws.
- **BUG (Vader): hostile `checkout_env` steers the API host.** WC select validation doesn't restrict POSTed values, and the mode splices into the hostname (`evil.example/` → attacker host) — a regression vs the old inert else-branch. **Fixed**: `ENVIRONMENT_MODES` allowlist (production/sandbox/staging); off-list → production (the pre-template host for every unrecognised value); garbage no longer preserved as a select option. Tested incl. injection strings.
- **Decision (Han): ABN production merchants' host changes** from `api.two.inc` to `api.achterafbetalen.abnamro.nl` once the overlay releases. **Intended** — Magento ABN overlay already ships the identical template (`magento-abn-plugin` `etc/brand.xml`), so ABN Magento production merchants already call that host. Parity, release-gated.
- **NITs (Leia/Han)**: comment truthfulness on explicit-Production, field default `'Production'`→`'PROD'`, test-host default derived from the brand template (no brand-domain literals left in base runtime code), test hygiene. All applied.
- **Accepted, documented**: (a) case-variant stored values (`sandbox`) now resolve to sandbox host where they used to hit production — legacy UI only ever wrote exact `PROD`/`SANDBOX`, wp-cli-seeded shops audited; (b) miscased `Sandbox` renders an unmatched select that flips to `PROD` on next save (Vader round-2 NIT) — requires hand-set miscased value; (c) dead merchant-hostname list removed (all four hosts confirmed retired); (d) malformed overlay template not guarded — trusted co-released code, pinned by the overlay's contract test.

**Round 2** (independent verifier): all round-1 findings confirmed fixed with file:line evidence, no new defects, 72 unit tests green. **CLEAN.**

Reviewed and posted by Claude.

PR #352, opened 2026-07-08
TWO-25041/feat: graceful invoice retrieval when order not yet fulfilled

dgjlindsay:
Implements [TWO-25041](https://linear.app/two-inc/issue/TWO-25041) (WooCommerce child of [TWO-25040](https://linear.app/two-inc/issue/TWO-25040)).

## Problem

The admin order page linked the invoice / credit-note buttons straight at the API PDF endpoint. When the order is still `FULFILLING`, the API answers `400 {"error_code": "ORDER_NOT_FULFILLED"}` and the merchant gets raw JSON in the browser — reads as a failure when it just means "not ready yet".

## Change

Route both buttons through a new plugin-owned admin-ajax handler (`wp_ajax_twoinc_download_invoice`) that fetches the PDF server-side via the existing `make_request` plumbing and implements the parent spec's state check:

- **Invoice fetch OK** → stream the PDF (content-type / `%PDF-` magic validated — never streams a JSON error body as a `.pdf`).
- **400 `ORDER_NOT_FULFILLED`** → `GET /v1/order/{id}` and branch on `state`:
  - `FULFILLING` → **info** notice: invoice not ready yet, try again later.
  - `FULFILLED` → retry the invoice fetch once; if it fails again, error notice as today.
  - any other state → **info** notice naming the state.
- **Any other error** → today's terminal-error behaviour, enriched with the API `error_code` where the response body carries one (`get_twoinc_error_msg` alone yields only the generic "code: 400" string for 4xx).

Non-stream outcomes park a one-shot transient notice and redirect back to the order edit screen (`$order->get_edit_order_url()` — HPOS-aware, unlike `get_edit_post_link` — passed through `esc_url_raw`), where an `admin_notices` hook renders it.

### Security / correctness details

- Capability gate is `edit_shop_orders` (the order-screen capability), **not** `manage_options` — shop managers can use this button today and must not be locked out.
- Nonce via `check_admin_referer('twoinc_admin_nonce')` on the `_wpnonce` query arg. Note: this is a deliberate departure from the existing handlers' `wp_verify_nonce($_POST['nonce'])` pattern, because this endpoint is a link-click navigation, not an XHR POST — called out in code comments.
- Two order id resolved server-side from order meta (legacy `tillit_order_id` fallback preserved); `variant` whitelisted to `original` | `credit_note`; `is_twoinc_order` re-asserted in the handler.
- The flow can chain up to three serial API calls on a blocking navigation, so these calls use a 10s timeout (scoped to this call site via `make_request`'s existing `$timeout` parameter; the 30s default is untouched for other call sites).

## Testing

`php tests/unit/run.php` — 77/77 pass (67 existing + 10 new specs): stream-through, FULFILLING info, FULFILLED retry-then-stream, FULFILLED retry-fail error (with `error_code` surfaced), other-state naming, non-ORDER_NOT_FULFILLED passthrough (no order-state fetch), missing order ref, 200-non-PDF guard, credit-note variant param, and the capability gate (deny without `edit_shop_orders`, pass with it).

---
*Implemented by Claude on Doug's behalf. Design + design review on the Linear ticket.*

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -131,6 +131,10 @@ if (!class_exists('WC_Twoinc')) {
                 // Add HTML in order edit page
                 add_action('woocommerce_admin_order_data_after_order_details', [$this, 'add_invoice_credit_note_urls']);
 
+                // One-shot notice from the invoice/credit-note download
+                // handler (ajax_download_invoice redirects back here)
+                add_action('admin_notices', [$this, 'render_invoice_download_notice']);
+
                 // Advanced Custom Fields plugin hides custom fields, we must display them
                 add_filter('acf/settings/remove_wp_meta_box', '__return_false');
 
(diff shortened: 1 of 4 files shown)


dgjlindsay:
## Self-review lane — adversarial round complete

**Round 1** (3 personas, parallel, against commit `cc017b6`):
- **Han (SRE)**: BUG — notice transient keyed by user-id only → cross-tab/cross-order notice misattribution when two orders' downloads are triggered close together. Timeout/API-down/streaming failure modes all checked clean.
- **Vader (adversarial QA)**: BUG (high) — blanket `edit_shop_orders` capability is an IDOR: any shop-manager-capable user can pull any order's invoice by ID, bypassing per-order scoping that multi-vendor/restricted-visibility plugins rely on. Also flagged nonce not scoped to order+variant. Response-trust, redirect safety, static-handler config context all checked clean.
- **Yoda (WP idiom)**: convergent on capability (should be per-object `edit_shop_order`, not blanket `edit_shop_orders`) and nonce scoping. Plus two cheap hardening NITs: sanitize order id before it lands in the `Content-Disposition` filename, and clear output buffers before streaming PDF bytes (known WP gotcha for corrupted binary downloads). HPOS compliance, i18n, escaping, nonce-mechanism choice all confirmed correct.

**Fix commit** `4d22973`: per-object capability check (`edit_shop_order`, checked after order load), nonce scoped to `order_id + variant` at both mint and verify sites, transient keyed by `user_id + order_id` with screen-aware render, filename sanitization, output-buffer flush before headers. 3 new tests added (80 total).

**Round 2** (independent verifier, opus): re-read the diff from scratch (not the fix commit's own description), re-ran the suite, checked for regressions the fix itself could introduce (capability-reorder edge cases, nonce mint/verify parity, screen-detection fail-closed behavior, test-stub honesty). All 5 fixes CONFIRMED against real code, 84/84 tests pass, `php -l` clean, no new bugs. Verdict: **SHIP**.

Merging via `--admin` per the plugin-repo self-review lane.

*Audit trail by Claude, on Doug's behalf.*

PR #362, opened 2026-07-10
feat: animated loading-dots on checkout term chips while fee quote is in flight

dgjlindsay:
## Context
Matches PrestaShop's chip behavior fix (same product decision): the checkout term chips must never leak the merchant's configured surcharge rate to the buyer — only the real quoted amount, or a loading indicator while the quote is pending. This repo's chips already did the safe thing by accident (blank until the fee-quote POST resolves) but had no loading indicator, unlike magento-plugin's animated three-dot chip loader.

## Fix
- \`twoincTermChips\` gains a \`feesLoaded\` flag. \`render()\` shows an animated three-dot \`.twoinc-term-chip__loading\` span per chip while unresolved, swapping to the real fee (or nothing, if ~0) once loaded — same as before, just no more bare-blank interim state.
- \`refresh()\`: dots reset on every checkout update (cart totals invalidate old quotes), settle on success, failure, or malformed response — never left spinning indefinitely, never falls back to a rate.
- When offset pricing is disabled / no fees URL configured, skips straight to the settled state (no dots that would never resolve).
- CSS timing copied from Magento's \`.two-term-chip__loading\` (1.4s staggered opacity pulse), adapted to this file's \`twoinc-\` naming and color conventions.

## Testing
\`node --check\` clean. No JS test harness in this repo — verified by manual trace of all four settle paths (success, malformed success, fail, no-fetch-configured).

---
*Reviewed by Claude prior to opening this PR — self-review-lane process.*

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -227,6 +227,37 @@
   color: inherit;
   opacity: 0.85;
 }
+/*
+ * Loading dots shown while the fee quote is in flight. Timing mirrors the
+ * Magento checkout (.two-term-chip__loading): 1.4s staggered opacity pulse.
+ */
+.twoinc-term-chip__loading {
+  display: inline-block;
+  letter-spacing: 2px;
+  font-size: 12px;
+  font-weight: 700;
+  color: inherit;
+}
+.twoinc-term-chip__loading > span {
+  display: inline-block;
+  animation: twoinc-term-chip-dot 1.4s infinite ease-in-out both;
+}
+.twoinc-term-chip__loading > span:nth-child(2) {
+  animation-delay: 0.2s;
+}
+.twoinc-term-chip__loading > span:nth-child(3) {
+  animation-delay: 0.4s;
+}
+@keyframes twoinc-term-chip-dot {
+  0%,
+  80%,
+  100% {
+    opacity: 0.2;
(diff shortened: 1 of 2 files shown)


dgjlindsay:
## Self-review summary

CI green (all 12 checks: unit tests 7.4-8.4, boot/upgrade smoke on WC 10.8.1/10.9.4, e2e, pre-commit). Gemini Code Assist hit its daily quota limit — no substantive review available.

**Own review:** confirmed the pre-existing chip logic never leaked a rate to the buyer (blank until fee-quote resolves) — this change only adds the missing loading-dots indicator for parity with magento-plugin's `.two-term-chip__loading`. Traced all four settle paths: success (dots → real amount), malformed/unsuccessful response (dots → blank, previously a silent no-op that would have left the dots stuck), fetch failure (same), and offset-pricing-disabled/no-fees-url (skips straight to settled state so dots never spin forever with no fetch to resolve them). `feesLoaded` resets on every `refresh()` call (debounced on WC's `updated_checkout` event only, not per-keystroke) since a cart-total change invalidates prior quotes — showing stale amounts would be worse than a brief dot flicker.

No blockers. Will merge after the standard 10-minute window from this comment with no new human comments.

*— by Claude*

PR #363, opened 2026-07-11
TWO-25070: configurable surcharge tax treatment

dgjlindsay:
## Summary
Adds a merchant-configurable surcharge tax treatment: Standard (default, unchanged), Specific tax class, or Never taxed. Closes the gap where the surcharge fee was hardcoded to WC's Standard tax class regardless of merchant intent (TWO-25069).

- New admin settings: `surcharge_tax_treatment` (standard/custom_class/always_zero), `surcharge_tax_class` (live-populated from `WC_Tax::get_tax_classes()`)
- `always_zero` uses `add_fee(..., taxable: false)`, not a binding to WC's "Zero rate" class (that class has no special semantics — a merchant could add rate rows to it later and silently break the guarantee)
- Deleted/stale tax class selection degrades explicitly to Standard (WC's `add_fee()` would otherwise silently revert to Standard with no signal) — surfaced via a visible admin notice
- Save-time validation enforces membership against the live class list

## Test plan
- [x] 86/86 unit tests pass (81 pre-existing regression, 5 new)
- [x] `php -l` / `node --check` clean
- [ ] Staging checkout spot-check with a multi-rate tax class (unit suite is stub-based, no live WC install in this repo)

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -406,6 +406,21 @@ jQuery(function ($) {
       $(".twoinc-surcharge-grid-field").toggle(type !== "none");
     }
 
+    // ── Surcharge tax treatment/class visibility ─────────────────────
+    // Both rows follow the surcharge method (no surcharge → no tax
+    // question); the class dropdown additionally follows the treatment
+    // (only "Specific tax class" needs it). Selection changes mirror
+    // without a save; the stored values are untouched until Save.
+    const $taxTreatment = $("#" + prefix + "surcharge_tax_treatment");
+    const $taxClass = $("#" + prefix + "surcharge_tax_class");
+
+    function updateTaxFields() {
+      const type = $surchargeType.val() || "none";
+      const treatment = $taxTreatment.val() || "standard";
+      $taxTreatment.closest("tr").toggle(type !== "none");
+      $taxClass.closest("tr").toggle(type !== "none" && treatment === "custom_class");
+    }
+
     function onTermsChanged() {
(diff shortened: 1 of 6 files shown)


dgjlindsay:
## Self-review summary

Reviewed directly (small, well-scoped diff) — no blockers found. 3-mode tax treatment (standard/custom_class/always_zero), live-validated against `WC_Tax::get_tax_classes()`, degrades safely on a deleted class (explicit fallback + admin warning, not silent), `always_zero` correctly uses `taxable=false` rather than binding to WC's semantics-free "Zero rate" class. 86/86 tests pass (81 pre-existing + 5 new).

Also fixed a pre-existing, unrelated bug this PR's CI run surfaced: `WC_Twoinc_Checkout.php`'s checkout-field priority filters assumed `billing_company` always exists in the fieldset, which isn't true when WooCommerce's own Company-name field is disabled — the resulting undefined-array-key warnings were leaking into the checkout AJAX response and breaking order placement (`e478615`, separate commit, unrelated to the surcharge tax-config change itself).

CI green. Starting the standard 10-minute quiet window before merge.

*— posted by Claude*

dgjlindsay:
Addressed Gemini's review: empty-slug guard on \`get_surcharge_tax_class_options()\` (prevents collision with the placeholder option) and HTML-escaping on the stale-class-notice interpolation (\`9aa0772\`). 86/86 tests pass, CI green. Restarting the 10-minute quiet window.

*— posted by Claude*

PR #366, opened 2026-07-15
TWO-25097/fix: fail loud on negative discount amounts

dgjlindsay:
## Summary

A negative line/order discount could reach the Two API silently — no sign check existed on any discount surface. This adds a fail-loud guard at all three discount surfaces in `class/WC_Twoinc_Helper.php`, mirroring the PrestaShop guard shipped under TWO-24741:

- **Per-line discount** in `get_line_items()` (`line_subtotal - line_total`)
- **Order-level discount** in `compose_twoinc_order()` (`get_total_discount()`)
- **Order-level discount** in `compose_twoinc_edit_order()` (same surface, edit PUT body)

New helper `guard_negative_discount()`: derives the discount at **native precision**, rounds **once** at the payload boundary, and on a genuinely negative value logs a diagnostic (surface, ids, operands) and throws — checkout fails loud, never silently clamps to 0.

## Rounding-order gotcha (from the PS round-1 self-review)

The sign check runs on the once-rounded value, never on early-rounded operands: rounding operands to 2dp before differencing manufactures phantom ±0.01 discounts and would fire the throw on legitimate carts. Sub-cent float residue in native totals rounds to zero and passes; negative-zero `-0.00` residue is normalised to `0.00` in the payload.

## Tests

Unit tests (custom runner, `make test-unit`) covering the ticket's (a)/(b)/(c):

- (a) legitimate positive discount passes through untouched (line + order surfaces)
- (b) genuinely negative discount throws with a clear message naming the product/order (all three surfaces)
- (c) rounding-order regression: `25.024 - 25.026` (native −0.002, early-rounded −0.01) does NOT false-positive and emits plain `0.00`

Test scaffolding: `StubProductLineItem` (ArrayAccess + `get_taxes()`, matching how `get_line_items()` reads WC_Order_Item_Product both ways) and `StubOrder::get_item_subtotal()`.

Ticket: [TWO-25097](https://linear.app/two-inc/issue/TWO-25097)

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -3597,18 +3597,39 @@ if (!class_exists('WC_Twoinc')) {
                 return false;
             }
 
-            $twoinc_order_hash = $order->get_meta(WC_Twoinc_Brand::meta_key('req_body_hash'));
-            $twoinc_updated_order_hash = WC_Twoinc_Helper::hash_order($order, $twoinc_meta);
-            $updated = true;
-            if (!$twoinc_order_hash || $twoinc_order_hash != $twoinc_updated_order_hash) {
-                $updated = $this->update_twoinc_order($order, $twoinc_meta);
-                if ($updated) {
-                    $order->update_meta_data(WC_Twoinc_Brand::meta_key('req_body_hash'), $twoinc_updated_order_hash);
-                    $order->save();
+            // Compose-time validation (e.g. the negative-discount guard,
+            // TWO-25097) throws to fail checkout loud. This path is fired
+            // from non-checkout hooks (admin order save, status
+            // transitions, fulfilment sync) — contain the failure to the
+            // Two sync: note + log it and abort the update, but do not
+            // crash the surrounding save/transition.
+            try {
+                $twoinc_order_hash = $order->get_meta(WC_Twoinc_Brand::meta_key('req_body_hash'));
+                $twoinc_updated_order_hash = WC_Twoinc_Helper::hash_order($order, $twoinc_meta);
+                $updated = true;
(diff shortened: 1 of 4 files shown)


dgjlindsay:
## Self-review summary

Adversarial review run in 2 rounds against the diff (`git diff origin/staging...HEAD`).

**Round 1 findings (1 blocking, 4 non-blocking) — all fixed:**
- 🔴 `get_line_items()` guard was also hit by `compose_twoinc_refund()`, whose line items carry negated amounts — a refund of a previously-discounted line would legitimately produce a negative diff and wrongly throw. Fixed with an `$is_refund` flag that bypasses the guard on that path (legacy `round_amt` behaviour preserved); covered by `testNegativeDiscountGuardSkipsRefundLineItems`.
- 🟡 The thrown exception carried raw operands/ids, which WooCommerce surfaces verbatim to the shopper as a checkout notice. Split into a short customer-safe `$subject` (product/order identifier) vs. a `$log_context` with the full diagnostic, log-only.
- 🟡 `compose_twoinc_edit_order`/`hash_order` also run from non-checkout hooks (admin order save, status transitions, fulfilment sync) via `process_update_twoinc_order`. An uncaught throw there would crash those flows rather than just the Two sync. Wrapped in try/catch → order note + log + abort sync, no crash.
- 🟡 Test gaps closed: refund-path case, positive order-level discount through both compose bodies.

**Round 2:** clean — no further findings. 3dp-precision behavior (a native diff representable at shop precision is treated as genuine data, not rounding residue) confirmed as intended design, not a gap.

CI: 12/12 checks green (unit-tests 7.4–8.4, pre-commit, boot smoke, upgrade smoke, e2e-tests).

— posted by Claude

PR #368, opened 2026-07-15
TWO-25104: FX conversion layer on /refdata/v1/fx-rates

dgjlindsay:
[TWO-25104](https://linear.app/two-inc/issue/TWO-25104) — WooCommerce half of the FX cutover (parent TWO-24815 / TWO-24739). Net-new capability: WooCommerce had **no FX layer at all** (the acknowledged single-currency-only parity gap vs Magento).

## What's here

**New `WC_Twoinc_FX`** — the plugin's single FX source, backed by checkout-api's live `GET /refdata/v1/fx-rates` (EUR-pivot spot table, merchant-API-key authed, TWO-24776):

- Full-table fetch via `make_request` (server-side only; key and rates never reach browser JS).
- **Cache**: request memo → 6h freshness transient → durable last-known-good option (non-autoloaded, survives transient expiry and fetch failures). Recurring 6h Action Scheduler refresh keeps it warm; deactivation unschedules it.
- **On-demand path**: a checkout in a currency the cached table lacks triggers one live re-fetch (5s timeout); a failed fetch arms a 5-min retry throttle so a flapping API is never hammered per conversion. (The ticket's "parallel with the order-intent call" is adapted: WC's checkout is synchronous PHP with no order-intent analogue, so it's one tight-timeout synchronous fetch at first use, then cached.)
- Cross rates computed exactly as the endpoint does (`eur_value(from) / eur_value(to)`); `as_of` stored for observability.

**Consumers wired, per ticket fail semantics:**

| Surface | Behaviour | On no rate |
|---|---|---|
| Platform + merchant min-order gates | basket converted into the minimum's currency, last-known-good rates | fail **closed** (only when no rate ever fetched) |
| Fixed surcharge + cap (`build_buyer_fee_share`) | converted store → active checkout currency before the pricing quote | surcharge **withheld** — a wrong-currency amount is never sent (fail-soft, checkout never blocked) |
| Cart fee (`apply_cart_fee`) | enters the basket at the pricing endpoint's output, no store re-conversion; a quote echoing a different currency than the cart is skipped | n/a |
| Settings help text (platform floor) | approximate converted figure appended | fail **soft** (native-only wording) |

The `WC_Twoinc_Payment_Terms` "no FX" acknowledgement comment is gone; same-currency stores never touch the FX layer (regression-pinned in tests).

**Behaviour change to note:** the merchant's own minimum previously failed *open* on a cross-currency basket (unjudgeable by design). It is now judged via FX and fails closed when no rate was ever fetched — uniform gate semantics per the ticket.

## Validation

- Unit suite (WC custom runner, php 8.2 container as CI runs it): 16 new tests — cache freshness/no-over-fetch and `as_of`, last-known-good fallback + retry throttle, malformed/non-2xx rejection, uncached-currency single re-fetch, cross-currency converted surcharge/cap and min-order gate outcomes from the endpoint rate, fail-closed/fail-soft matrix, same-currency no-op pins. All green.
- Manual staging QA (display currency ≠ merchant base currency: converted fixed surcharge + min-order gate) still owed before this is called done — tracked on the ticket.

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -2287,26 +2287,36 @@ if (!class_exists('WC_Twoinc')) {
             $basket_is_judgeable = !WC()->cart->is_empty()
                 && !(function_exists('is_wc_endpoint_url') && is_wc_endpoint_url('order-pay'));
 
+            // A basket in another currency is judged by converting its
+            // value into the minimum's currency via the Two FX layer
+            // (TWO-25104): last-known-good rates, so a transient rates-API
+            // failure never flaps the gateway. Null (no rate ever fetched,
+            // or an unsupported currency) fails CLOSED — a basket that
+            // cannot be proven to satisfy a minimum is not offered the
+            // gateway; the API still enforces the platform minimum at
+            // order creation. Same-currency baskets never touch the FX
+            // layer.
+            $meets_minimum = function (array $minimum) use ($basket_value): bool {
+                $value = $basket_value($minimum['basis']);
+                $basket_currency = get_woocommerce_currency();
+                if ($basket_currency !== $minimum['currency']) {
+                    $value = WC_Twoinc_FX::convert($this, $value, $basket_currency, $minimum['currency']);
+                    if ($value === null) {
+                        return false;
+                    }
+                }
+                return $value >= $minimum['amount'];
+            };
+
             $satisfied = true;
             if ($platform_minimum && $basket_is_judgeable) {
-                // Fail closed on a basket in another currency: with no FX
-                // source in WooCommerce it cannot be proven to satisfy the
-                // funding partner's minimum.
-                $satisfied = get_woocommerce_currency() === $platform_minimum['currency']
-                    && $basket_value($platform_minimum['basis']) >= $platform_minimum['amount'];
+                $satisfied = $meets_minimum($platform_minimum);
             }
             if ($satisfied && $gate) {
                 $satisfied = in_array(WC()->customer->get_billing_country(), $gate['billing_countries'], true);
             }
             if ($satisfied && $merchant_minimum && $basket_is_judgeable) {
-                $satisfied = get_woocommerce_currency() === $merchant_minimum['currency']
-                    ? $basket_value($merchant_minimum['basis']) >= $merchant_minimum['amount']
-                    // The merchant minimum is store-currency scoped; a
-                    // basket in another currency cannot be judged against
-                    // it (no FX source in WooCommerce) — fail open on the
-                    // merchant's own optional bar, the platform minimum
-                    // above still applies.
-                    : $satisfied;
+                $satisfied = $meets_minimum($merchant_minimum);
             }
             if (!$satisfied) {
                 unset($available_gateways[$this->id]);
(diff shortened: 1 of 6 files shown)


dgjlindsay:
## Self-review summary — by Claude

Adversarial 4-persona review (Leia dropped, Han/Vader/Yoda + **Obi-Wan** added given payment-amount and currency-handling risk), 2 rounds against the worktree diff.

### Round 1 — findings and fixes
Convergent (3 reviewers: Han, Obi-Wan, Vader): a checkout in a currency the cached table lacked was re-fetching **every single request**, forever, even when the table was already fresh (and thus conclusive) — an unbounded synchronous fetch loop disguised as a cache, with real checkout-latency and API-hammering impact. **Fixed**: the on-demand refetch now only fires when the working table is genuinely stale.

Convergent (2 reviewers: Han, Yoda): freshness TTL == Action Scheduler refresh cadence with zero slack — AS is best-effort, so checkout traffic routinely paid for the fetch right at the 6h boundary. **Fixed**: added a `FRESHNESS_GRACE` margin plus a durable `fetched_at`-based freshness fallback (covers object-cache eviction of the transient), and added the missing `$unique=true` on `as_schedule_recurring_action` to close a duplicate-job race on cold start.

Reproduced-and-fixed (verified empirically, 1 reviewer each):
- Vader: a stored FX table reaching `wp_options` via any route other than the plugin's own fetch (DB import, plugin-version skew, manual edit) wasn't re-validated on read → a poisoned rate hit the division in `get_rate()` → **`DivisionByZeroError` fatal on the checkout path**. Fixed by sharing the same rate-sanitization filter between fetch and read paths.
- Han: a surcharge cap that rounded to 0.00 after conversion silently dropped only the cap, leaving the percentage fee **uncapped** — the opposite of the merchant's configuration. Fixed to withhold the whole surcharge in that case, consistent with the no-rate-available path.
- Vader: the cart-fee currency-mismatch guard wasn't normalized (case/whitespace) like the rest of the FX layer. Fixed.
- Yoda: the FX recurring-refresh hook registration at `plugins_loaded` forced early materialization (and permanent caching) of the brand config, ahead of the documented window a same-priority brand overlay has to register first. Fixed by deferring both registrations to `init`.
- Obi-Wan: API response `base` field wasn't type-checked before use. Fixed.

7 new/updated unit tests landed with round 1 (24 FX tests total); full suite green in the CI `php:8.2` container both before and after.

### Round 2 — clean
All four personas re-reviewed the fixed code independently (not just the round-1 summary) — hand-traced the logic, tried to break each fix (array/object rates, missing/non-numeric `fetched_at`, case/whitespace variants, hook-timing races), and re-ran the exploit scenarios round 1 found. **Zero new BLOCKER/BUG findings.** Landed two trivial items surfaced (missing `/* translators: */` comment; a docstring accuracy tweak on which code path actually performs the on-demand refetch in practice). Two OBSERVATIONs were explicitly accepted as out of scope for this round: a bounded, self-healing cache-stampede window on concurrent expiry (Han — architectural, not introduced by this PR's fixes), and the accepted fail-soft-to-no-surcharge tradeoff extending slightly further given the widened freshness window (Obi-Wan — same direction as the already-locked design decision, not a new exploit).

One intentional behavior change flagged by Obi-Wan and worth calling out explicitly: the merchant's own (optional) minimum-order value previously failed **open** on a cross-currency basket (unjudgeable, so it didn't gate). It now fails **closed** like the platform minimum, consistent with the ticket's uniform gate semantics — this changes gateway availability for existing multi-currency merchants using a merchant minimum, not just new deployments.

CI green (unit tests 7.4–8.4, boot smoke, upgrade smoke, e2e, pre-commit) before and after both rounds of fixes.

PR #372, opened 2026-07-27
TWO-24816/fix: source the API locale from determine_locale()

dgjlindsay:
## Audit (TWO-24816)

`get_locale()` in this plugin is a helper, not WordPress's `get_locale()`. Every call site goes through it — there are no direct WordPress locale calls anywhere in the plugin:

| Call site | Locale used for | Verdict |
| --- | --- | --- |
| `class/WC_Twoinc_Helper.php:1076` (`WC_Twoinc_Helper::get_locale`) | The single source for both consumers below | Was `get_user_locale()` — wrong on the storefront, fixed here |
| `class/WC_Twoinc.php:3774` (`make_request`) | `Accept-Language` header on every API request, including checkout order create/update | Buyer/request locale is what is meant; `determine_locale()` is the WordPress function that resolves it |
| `class/WC_Twoinc.php:1507` (`resolve_invoice_download`) | `lang` query parameter on the invoice PDF fetch | Admin-ajax, so the admin user's locale — behaviour unchanged (`determine_locale()` returns the user locale in admin) |
| `snippets/two-demo/functions-part.php:19` | Picks a hard-coded checkout field label per language, demo snippet only | Storefront locale is what is meant; inherits the helper fix, snippet untouched |

## Fix

Only the helper changes: `determine_locale()` instead of `get_user_locale()`.

On the storefront the page — including this plugin's own translated strings — is rendered in the site or language-switched locale, which WordPress resolves with `determine_locale()`. `get_user_locale()` instead returns a logged-in buyer's WordPress profile language, so a buyer whose profile is set to one language while shopping the store in another had the API answer in a different language from the page around it. In wp-admin and admin-ajax `determine_locale()` resolves to the user locale anyway, so the invoice-download path is unchanged.

### Deliberately not changed

- **The underscore form (`nb_NO`, not `nb-NO`).** The API matches the `lang` query parameter literally against an allow-list of underscore locales, so hyphenating it would break the invoice PDF language. On the header side Werkzeug normalises the separator before matching, so `nb_NO` resolves correctly there too. Documented on the helper so it does not get "corrected" later.
- **No translation-loading changes, no user-visible strings.**

## Test coverage

`testLocaleFollowsRequestLocaleWithEnglishFallback` covers both branches (resolved locale, and the `en_US` fallback on an empty result). The harness stub for `get_user_locale()` is replaced by one for `determine_locale()`, so a regression back to the user locale fails loudly with an undefined function rather than silently.

## Second commit: red base

Staging is currently red on both static-analysis gates, unrelated to this work — TWO-25108 added the gates and TWO-25104 added the FX layer, both green when they last ran, but neither re-ran after the other merged. `TWO-25104/fix` in this PR carries the fix (one PSR-12 multi-line condition, one mis-indented pair of statements in a test closure, and eleven phpstan `property.notFound` errors caused by `fxGateway()`'s return type widening away the anonymous class that declares `$fx_requests`). The fix is only in this PR — it is not on staging.

Local runs of the CI commands, on this branch:

- `make phpcs` → exit 0 (warnings only; `ignore_warnings_on_exit`)
- `make phpstan` → `[OK] No errors`
- `make test-unit` → `All tests passed.`

### DIFF
--- a/class/WC_Twoinc_FX.php
+++ b/class/WC_Twoinc_FX.php
@@ -353,7 +353,8 @@ if (!class_exists('WC_Twoinc_FX')) {
             }
             $from_value = self::base_value($table, $from);
             $to_value = self::base_value($table, $to);
-            if (($from_value === null || $to_value === null)
+            if (
+                ($from_value === null || $to_value === null)
                 && $gateway
                 && !self::$fetched_this_request
                 && !self::is_fresh($table)
--- a/class/WC_Twoinc_Helper.php
+++ b/class/WC_Twoinc_Helper.php
@@ -1067,15 +1067,25 @@ if (!class_exists('WC_Twoinc_Helper')) {
         }
 
         /**
-         * Get the user locale in full form, e.g. en_US — sent as the
-         * invoice PDF `lang` parameter and the Accept-Language header,
-         * both of which accept the full form.
+         * Get the locale of the current request in full form, e.g. en_US —
+         * sent as the invoice PDF `lang` parameter and the Accept-Language
+         * header, both of which accept the full form (underscore, not
+         * hyphen: the API matches `lang` literally against its allow-list).
+         *
+         * determine_locale(), not get_user_locale(): on the storefront the
+         * page — including this plugin's own translated strings — is rendered
+         * in the site/switched locale, so that is the language the API should
+         * answer in too. get_user_locale() instead returns the WP profile
+         * language of a logged-in buyer, which can differ from the checkout
+         * page around it. In wp-admin (and admin-ajax, i.e. the invoice
+         * download) determine_locale() resolves to the user locale anyway,
+         * so that path is unchanged.
          *
          * @return string
          */
         public static function get_locale()
         {
-            $locale = get_user_locale();
+            $locale = determine_locale();
             if ($locale && strlen($locale) > 0) {
                 return $locale;
             }
--- a/tests/unit/bootstrap.php
+++ b/tests/unit/bootstrap.php
@@ -103,9 +103,9 @@ function wc_get_price_decimals()
     return 2;
 }
 
-function get_user_locale()
+function determine_locale()
 {
-    return 'en_US';
+    return $GLOBALS['__twoinc_test_locale'] ?? 'en_US';
 }
 
 function is_admin()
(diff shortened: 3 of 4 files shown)


dgjlindsay:
## Self-review summary

Adversarial review of `git diff origin/staging...HEAD` — **one clean round, no findings**.

What was checked, and why each concern came up clean:

- **`determine_locale()` availability.** It landed in WordPress 5.0, and `readme.txt` declares `Requires at least: 6`, so there is no fatal-on-old-WP path.
- **Stub rename, not addition.** `tests/unit/bootstrap.php` renames the `get_user_locale()` stub to `determine_locale()`. No remaining caller of `get_user_locale()` exists in the plugin or the harness, so nothing is left pointing at a function the harness no longer defines. (The new test comments on exactly this: reverting the production call without reverting the stub would fatal in CI, which is a useful tripwire.)
- **Request-lifecycle safety.** Every caller of `WC_Twoinc_Helper::get_locale()` runs well after `plugins_loaded`/`init` — checkout rendering and the admin-ajax invoice download — so the locale filters `determine_locale()` consults are registered by the time it is called.
- **Behavioural intent.** On the storefront the surrounding page, including this plugin's own translated strings, is rendered in the site/switched locale, so that is the language the API should answer in. In wp-admin and admin-ajax `determine_locale()` resolves to the user locale anyway, so the invoice-download path is unchanged.
- **Wire format.** Still the full underscore form (`en_US`), which is what the API matches `lang` against and what `Accept-Language` accepts here; the empty-result fallback to `en_US` is covered by the new test.

### Second commit: the static-analysis gate repair

`origin/staging` is red on both gates independently of this branch — the TWO-25108 gate PR and the TWO-25104 FX PR were each green when last run, neither re-ran after the other merged, and the workflows are `pull_request`-only so nothing runs against `staging` itself. This branch carries the repair so it can be reviewed green.

woocommerce-plugin PR #371 carries its own copy of the same repair. The two are byte-identical in code; they differ only in where the `fxGateway()` rationale comment sits. This PR keeps it inside the existing docblock rather than as a loose `//` block between the docblock and the signature, so this is the copy being landed, and woocommerce-plugin PR #371 will be rebased with its redundant repair commit dropped.

Dropping `fxGateway()`'s `: WC_Twoinc` return type is the right fix rather than a suppression: the declared type widens the returned anonymous subclass away, which is precisely why phpstan could not see the `$fx_requests` counter the FX tests assert on. No caller relies on the declared type.

*Review by Claude.*

PR #373, opened 2026-07-27
TWO-25170/fix: resolve checkout page host from the same environment as the API host

dgjlindsay:
## Problem

The plugin resolves the API host and the hosted checkout-page host through two different code paths, and only the API path honours the legacy dev-host sniff:

- `WC_Twoinc::get_twoinc_checkout_host()` (`class/WC_Twoinc.php`) — when the stored mode resolves to `production` **and** `WC_Twoinc_Helper::is_twoinc_development()` matches the shop hostname, it returns the `test_checkout_host` option, i.e. a staging API host.
- `WC_Twoinc_Sole_Trader::get_signup_page_url()` → `WC_Twoinc_Helper::get_environment_host('checkout', …)` — no dev branch, re-derived from the stored mode alone, so it returned the **production** checkout host.

On a staging shop with `checkout_env` never configured, the browser was therefore told: mint against staging, open the popup on production. The production signup app validates the delegation token against the production API, which rejects it (`401 AUTHENTICATION_INVALID`), and the sole-trader flow dies inside the popup. That is the reported TWO-25170 symptom.

Magento and PrestaShop do not have this bug because both derive the API host and the page host from the same setting. WooCommerce was the only one with an escape hatch on one host and not the other.

## Fix

`WC_Twoinc_Helper::get_effective_environment_mode()` — the environment the gateway *actually* talks to. It equals the stored mode except on a dev-sniffed shop carrying the never-configured default, where the API host comes from the free-text `test_checkout_host` option rather than the brand template; there the mode is read back out of that host.

`get_environment_host()` now resolves off the effective mode, so every service host the gateway emits sits in the same environment as its API host. That is the invariant the bug violated, and it means a future page-host consumer inherits correct behaviour instead of needing its own special case — no special-casing of the sole-trader URL.

Classification is template-driven off the brand registry's `checkout_url_template`, with no brand-specific branch. A test host the classifier cannot place (localhost, a bespoke tunnel) falls back to `staging`: a dev-sniffed shop is by definition not a production shop, `staging` is the option's own default, and a test environment can neither take real money nor accept a production token.

## Brand overlays

`woocommerce-abn-plugin` supplies only `checkout_url_template` and overrides none of `get_environment_host()`, `get_signup_page_url()` or the `twoinc_sole_trader_signup_url` filter — checked by grepping that repo's `staging` branch for all of them. It inherits the fix unchanged, and the new test covers an overlay brand explicitly (`api.staging.<brand>` → `checkout.staging.<brand>`). No change needed in the overlay repo.

## Tests

New `testServiceHostsShareTheApiHostsEnvironment` asserts the API host and the checkout-page host agree on environment across: dev-sniffed shop on a staging test host, on a sandbox test host, on an unclassifiable test host, on a deliberately production test host, a real merchant shop, and an overlay brand. Existing environment-mode/host tests are unchanged and still pass.

```
$ php tests/unit/run.php
PASS BrandConfigSpec::testSoleTraderSignupUrlFollowsEnvAndFilter
PASS BrandConfigSpec::testEnvironmentHostFollowsModeAndBrandTemplate
PASS BrandConfigSpec::testCheckoutHostPrefersExplicitModeOverDevSniffing
PASS BrandConfigSpec::testServiceHostsShareTheApiHostsEnvironment
...
All tests passed.
```

## Static analysis is red on the base branch

`staging` is already failing both static-analysis gates — **phpcs 3 errors** (`class/WC_Twoinc_FX.php:356`, `tests/unit/run.php:974-975`) and **phpstan 11 errors** (all `WC_Twoinc::$fx_requests` in `tests/unit/run.php`) — pre-existing, none in code this PR touches. Running both pinned tools (phpcs 4.0.1, phpstan 2.2.5, same stub SHAs as CI) with and without this change gives byte-identical findings; phpstan's only difference is line numbers shifting past the inserted test. So a red static-analysis job here is base breakage, not this PR. The gate repair is already carried by woocommerce-plugin PR #371 and PR #372, so it is deliberately not duplicated here.

## Not done here

The legacy hostname sniff arguably ought to be retired in favour of an explicit `checkout_env` everywhere, rather than mirrored into a second resolution path. That is a wider and riskier change — it silently repoints any dev install that never set `checkout_env` — and is out of scope for an urgent fix.

Linear: TWO-25170

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -197,6 +197,10 @@ if (!class_exists('WC_Twoinc')) {
             // an explicit Production selection is indistinguishable from the
             // never-configured default, so the sniffer still applies there.
             // Set checkout_env instead of extending the sniffer.
+            //
+            // WC_Twoinc_Helper::get_effective_environment_mode() encodes the
+            // same condition so that every other host the gateway emits lands
+            // in the environment this host picks; keep the two in step.
             if (
                 WC_Twoinc_Helper::get_environment_mode($this) === 'production'
                 && WC_Twoinc_Helper::is_twoinc_development()
--- a/class/WC_Twoinc_Helper.php
+++ b/class/WC_Twoinc_Helper.php
@@ -1048,11 +1048,79 @@ if (!class_exists('WC_Twoinc_Helper')) {
         }
 
         /**
+         * The environment the gateway actually talks to, i.e. the one the API
+         * host resolves to — which is not always the configured mode.
+         *
+         * On a dev-sniffed shop (see is_twoinc_development()) carrying the
+         * never-configured default mode, the API host comes from the free-text
+         * `test_checkout_host` option rather than from the brand template, so
+         * the configured mode ('production') is not the environment in use.
+         * Every host the gateway hands the browser has to follow the API host,
+         * or the shop mints tokens in one environment and sends the buyer to
+         * an app in another (TWO-25170: staging-minted delegation tokens
+         * rejected 401 by the production API behind the production hosted
+         * signup page). So read the environment back out of the test host.
+         *
+         * @param WC_Payment_Gateway $gateway
+         *
+         * @return string one of ENVIRONMENT_MODES
+         */
+        public static function get_effective_environment_mode($gateway)
+        {
+            $mode = self::get_environment_mode($gateway);
+            if ($mode !== 'production' || !self::is_twoinc_development()) {
+                return $mode;
+            }
+            return self::environment_mode_of_host((string) $gateway->get_option('test_checkout_host'));
+        }
+
+        /**
+         * Classify an API host into an environment mode: the brand's
+         * production API host is 'production', a host whose leading labels are
+         * `api.<mode>` is that mode (any domain — dev shops legitimately point
+         * the test host off the brand's domains).
+         *
+         * A dev-sniffed shop is by definition not a production shop, so a host
+         * this cannot classify (localhost, a bespoke tunnel) falls back to
+         * 'staging' — the option's own default, and the safe direction: a test
+         * environment can neither take real money nor accept a production
+         * token.
+         *
+         * @param string $host
+         *
+         * @return string one of ENVIRONMENT_MODES
+         */
+        private static function environment_mode_of_host($host)
+        {
+            $hostname = (string) parse_url($host, PHP_URL_HOST);
+            $production = (string) parse_url(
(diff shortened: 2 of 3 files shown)


dgjlindsay:
## Rebased, and self-review summary

**Rebase:** this branch has been rebased onto the current `staging`. Its phpcs and phpstan jobs were red purely from base breakage — `staging` was carrying three phpcs errors and eleven phpstan `property.notFound` errors, because the TWO-25108 gate PR and the TWO-25104 FX PR were each green when last run and neither re-ran after the other merged, and the workflows are `pull_request`-only so nothing runs against `staging` itself. That repair has since landed on `staging` (woocommerce-plugin PR #372). Verified on the rebased branch with the pinned tool versions the workflow uses — phpcs exit 0, phpstan `[OK] No errors`.

**Adversarial review of `git diff origin/staging...HEAD` — one clean round, no findings.** The concern worth recording is blast radius, since this changes host resolution:

- **Can a real merchant shop be flipped to staging hosts?** No. `get_effective_environment_mode()` only diverges from the configured mode when *both* the mode is the never-configured `production` default *and* `is_twoinc_development()` is true. That sniffer matches the home URL against localhost, the `TWOINC_DEV_HOSTNAMES` allow-list, and a fixed set of internal `*.two.inc` sub-domains; a real merchant domain cannot match it. The new test pins that case explicitly.
- **Which hosts actually move?** Only `get_environment_host('checkout', …)` for dev-sniffed shops — in practice the sole-trader hosted signup page. The `'api'` resolution is only reached down the non-dev branch, where effective mode equals configured mode by construction, so no API host changes.
- **Unclassifiable test hosts.** `parse_url()` yielding no host (a bare hostname, a bespoke tunnel, `localhost:8080`) falls through to `'staging'`. That is the option's own default and the safe direction — a test environment can neither take real money nor honour a production token — and it is covered by a test case.
- **Mode allow-list integrity.** `get_environment_mode()` already filters its return through `ENVIRONMENT_MODES` on exit, and `environment_mode_of_host()` can only return `'production'`, `'staging'`, or a label it has just checked against `ENVIRONMENT_MODES`. The `$labels[1] !== 'production'` guard is what stops a host like `api.production.<domain>` being read as a mode suffix; the brand's real production host is matched exactly, ahead of the suffix branch.
- **Brand-agnostic.** Resolution stays template-driven off the brand registry, with no brand-specific branch, and the test exercises an overlay's own domains.

*Review by Claude.*

PR #390, opened 2026-07-28
TWO-25230/chore: Seed .next-major escape hatch at 2

dgjlindsay:
## What

New root `.next-major` file containing `2`.

## Why

`.github/scripts/decide-bump-level.sh` implements a major-version escape hatch with two independent signals: a **declared** target major read from a root `.next-major` file, and a **discovered** breaking change (`feat!:` / `BREAKING CHANGE:` footer). The higher wins, and a major only fires when the target is above the current major.

Until now the declared half had no file to read in any of the plugin repos, so the mechanism was invisible unless you went and read the bump script. Seeding the file at the current major (2) makes it:

- **inert** — `target == current_major`, so the branch rule (staging -> patch, main -> minor) still applies unchanged;
- **discoverable** — a planned major becomes a one-character edit in a reviewable PR.

Format matches what the script parses: first whitespace-delimited token is the target major as a bare integer, with any remainder on the same line treated as a human-readable reason.

CI is deliberately never allowed to clear the file — the `target > current_major` condition disarms it on its own once the declared major has shipped. A declaration that has fallen *below* the current major is a hard failure, which is why this is seeded at 2 and not at 3.

Same change is going into all six plugin repos.


dgjlindsay:
## Self-review (adversarial pass)

One round, no blocking findings.

- **Format matches the parser.** `decide-bump-level.sh` reads `head -1` of `.next-major`, takes the first whitespace-delimited token via `awk`, and rejects anything non-integer. A bare `2` with a trailing newline parses cleanly; the remainder of the line (empty here) is the optional human reason.
- **Behaviourally inert.** With the current major at 2, `declared = 2` gives `target = max(declared, discovered) = 2`, which is not `> current_major`, so control falls through to the branch rule unchanged (`staging` -> patch, `main` -> minor). Verified by running the script on this branch: `level=patch`, `reason=branch rule: staging -> patch`.
- **Not a stale declaration.** The one state the script treats as a hard error is `declared < current_major`; seeding at exactly the current major cannot hit that, and it self-disarms rather than needing CI to clear it.
- **No other consumer.** Nothing else in the repo reads `.next-major`, so there is no second parser to keep in sync.
- **No packaging impact.** The file sits at the repo root and is not referenced by any build, release or distribution step.

CI green.

---
*by Claude*

PR #391, opened 2026-07-28
TWO-25232/fix: bound company search and split failure from empty

dgjlindsay:
## Why

Company search had **no client-side timeout at all**, so a request that never completed left the dropdown spinning forever. Worse, `language.errorLoading` deliberately returned the "searching…" string, so every real failure was indistinguishable from an in-flight search — and once the widget re-rendered, from "no matches". A buyer whose lookup fails is told, in effect, that their company does not exist.

## What

- **30s request timeout** (`companySearchTimeoutMs`). Deliberately wider than the backend's own retry envelope for the upstream provider lookup, so a slow-but-arriving response is never cut off client-side; this is the backstop for a request that does not arrive at all.
- **Custom select2 ajax `transport`.** select2's stock failure handler cannot tell a cancelled request from a timed-out one — it treats any `jqXHR` with `status === 0` as a cancellation, and a jQuery timeout also reports `status 0`. So the transport owns the messaging: `textStatus === "abort"` stays **silent** (select2 aborts the in-flight search on every keystroke, and the widget is re-created on country change), everything else raises "Company search is temporarily unavailable. Please try again."
- **`degraded` response flag consumed.** An HTTP 200 whose near-empty result set is unreliable because the upstream lookup timed out gets the same unavailable message rather than reading as "no companies found". Coded defensively — `data.degraded === true`, so an absent field (backend not yet deployed) reads as not degraded. `processResults` also no longer assumes `response.items` is an array.
- **Spinner inside the search field**, reusing the one loading idiom in this plugin (`.twoinc-dots`, shared with the order-intent check and the term-chip fee quote) rather than inventing a second one. Cleared on every outcome via `always()`.
- **Debounce 200ms → 300ms**, aligning all three plugin checkouts.
- New localised string `text.company_search_unavailable`, with a hardcoded fallback in JS for a page served by older cached PHP.

## Verification

- `phpcs` (4.0.1, `phpcs.xml`) — 0 errors.
- `phpstan` (2.2.5, level 2, with the pinned WP/WC stubs) — `[OK] No errors`.
- `php tests/unit/run.php` — all tests passed.
- `prettier@3.1.0 --check` on both touched assets — clean.
- **There is no JS test harness in this repo** (no `package.json` at the root, no unit coverage for `assets/js/`), so the JS changes carry no automated test. Stating that plainly rather than inventing a harness.
- **No browser was available: the spinner and the unavailable-message UX are visually unverified.** On the staging shop, worth looking at (a) the three-dot pulse appearing at the right-hand edge *inside* the company-search field while a query is in flight, and not clipping or overlapping typed text, (b) throttling/blocking `/companies/v2/company` to confirm the unavailable message replaces the results list instead of "no matches", and (c) typing quickly to confirm cancelled requests still produce no message at all.

## Overlay

The private brand overlay does **not** mirror any of this — its JS touches only the terms checkbox, its CSS has no select2 or loader rules, and it does not override the checkout text map. **No companion PR needed.**

## Adjacent pre-existing bugs — verified still on `staging`, NOT fixed here

Left alone as out of scope; filing/fixing is a separate call:

1. `assets/js/twoinc.js:1658` — the `#search_company_btn` handler restores `window.twoinc.enable_company_search` to a hardcoded `"yes"`, ignoring the merchant setting.
2. `twoincDomHelper.clearSelectedCompany()` (`:516`) re-wires `selectWoo` unconditionally on country change, with no `enable_company_search` gate — so it can resurrect the widget on a shop where search is off.
3. `assets/js/twoinc.js:1922` — `invalidFields.append("billing_phone_field")` calls `.append()` on a plain array; the invalid-phone path throws `TypeError`.

Ticket: [TWO-25232](https://linear.app/two-inc/issue/TWO-25232)

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -272,6 +272,32 @@
   }
 }
 
+/*
+ * Company-search spinner (TWO-25232). Sits inside the select2 search field
+ * itself, not in the results list, so it reads as "this field is working"
+ * and does not fight the results message it would otherwise replace. Same
+ * .twoinc-dots pulse as every other loading state here. Only laid out while
+ * .twoinc-searching is set, so the reserved right-hand gutter appears with
+ * the spinner instead of permanently narrowing the input.
+ */
+.select2-search--dropdown.twoinc-searching {
+  position: relative;
+}
+.twoinc-search-spinner {
+  display: none;
+}
+.select2-search--dropdown.twoinc-searching .twoinc-search-spinner {
+  display: inline-block;
+  position: absolute;
+  top: 50%;
+  right: 14px;
+  transform: translateY(-50%);
+  pointer-events: none;
+}
+.select2-search--dropdown.twoinc-searching .select2-search__field {
+  padding-right: 30px;
+}
+
 /* Sole trader toggle (TWO-24754) */
 /*
  * Sole-trader mode chips + signup note. Palette mirrors the Magento
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -43,6 +43,64 @@ let twoincUtilHelper = {
 
 let twoincSelectWooHelper = {
   /**
+   * Hard ceiling on a single company-search request, ms (TWO-25232). Before
+   * this there was no client timeout at all, so a request that never
+   * completed left the dropdown spinning forever. Deliberately wider than
+   * the backend's own retry envelope for the upstream provider lookup, so a
+   * slow-but-arriving response is never cut off client-side — this is the
+   * backstop for a request that does not arrive at all.
+   */
+  companySearchTimeoutMs: 30000,
+
+  /**
+   * Text for a company search that could not be completed. Read lazily
+   * because window.twoinc is populated by the checkout render; the literal
+   * is the last-resort fallback for a page where the localised string is
+   * missing (older cached PHP, brand overlay that trims the text map).
+   */
+  companySearchUnavailableText: function () {
+    return (
+      (window.twoinc && window.twoinc.text && window.twoinc.text.company_search_unavailable) ||
+      "Company search is temporarily unavailable. Please try again."
+    );
+  },
+
+  /**
+   * Toggle the in-field search spinner. Reuses the one loading idiom in
+   * this plugin (.twoinc-dots — shared with the order-intent check and the
+   * term-chip fee quote) rather than introducing a second one.
+   *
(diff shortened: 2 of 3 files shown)


dgjlindsay:
## Self-review (adversarial, 3 rounds)

Reviewed my own diff against select2 4.x's ajax-adapter contract, the dropdown-search markup, and the failure/abort/degraded state matrix. Two real findings, both fixed on this PR; third round clean.

**Round 1 — spinner race (fixed, `8e05f49`).** The spinner and the unavailable message are shared state keyed on nothing. A superseded request's `always()` could hide the spinner while a newer request was still in flight, and a superseded failure could paint an error over fresher results. Today select2 aborts the previous request *before* dispatching the next, so the hide always lands before the next show and the bug does not fire — but that is an internal detail of its ajax adapter, and the failure mode (permanently hidden spinner) is silent. Added a `companySearchSeq` supersession guard rather than leaning on the ordering.

**Round 1 — transport callback/promise shape (not a defect).** Raised as a possible contract mismatch: the transport both calls `success(data)` and returns the jqXHR. Checked against select2 4.x's own default transport, which is exactly `$req = $.ajax(params); $req.then(success); $req.fail(failure); return $req;` — same shape, and the returned object is required so select2 can `.abort()` it. The one intentional divergence is that `failure()` is never called: select2's failure handler treats any `status === 0` jqXHR as a cancellation, and a jQuery timeout also reports `status 0`, so it cannot make the distinction this ticket is about.

**Round 2 — guard incomplete (fixed, `6de1497`).** The guard covered `fail()` and `always()` but not `done()`, so a stale response could still repopulate the list under a newer search. Applied to all three.

**Round 3 — clean.** No remaining defects: degraded path (`success({items: []})` then `results:message`) leaves the dropdown in a consistent state, `abort` correctly stays silent, `processResults` no longer assumes `response.items` is an array, and the spinner insert is idempotent and no-ops when the dropdown is closed.

### Not covered by review, needs eyes

- **No JS test harness exists in this repo**, so none of the above is regression-protected by a test. The green checks cover PHP only (`phpcs`, `phpstan`, unit matrix, boot/upgrade smoke, e2e).
- **No browser was available — the spinner and the unavailable message are visually unverified.** See the verification section of the PR description for the three things to look at on the staging shop.

### Deliberately out of scope

Three pre-existing bugs in the same file, all verified still present on `staging`, listed in the PR description and left alone.

PR #392, opened 2026-07-29
TWO-25244/test: Jest harness + company-search regression tests

dgjlindsay:
## Why

The plugin had no JS test setup at all — no `package.json`, no jest, no CI gate — while `magento-plugin` (`Test/Js/`) and `prestashop-plugin` (`tests/js/`) both have one. Every company-search defect fixed under TWO-25232 was invisible to the PHP suite, because none of them lives in PHP.

Closes TWO-25244.

## What

Mirrors `prestashop-plugin`'s layout, which was built to the same brief:

| | |
| -- | -- |
| `package.json` (new) | root manifest, `private: true`, JS devDeps only, `test:js` script |
| `tests/js/jest.config.js` | beside the tests, `rootDir` back at the repo root, jsdom |
| `tests/js/wc-harness.js` | stands up the real browser |
| `tests/js/*.test.js` | 59 tests over `assets/js/twoinc.js` |
| `tests/js/README.md` | what is covered, what is not, and why the harness looks like this |
| `.github/workflows/unit-tests.yaml` | new `Jest (Node 20)` job — a **real** gate, no `continue-on-error` |
| `Makefile` | `make test-js` |
| `.distignore` | none of it reaches the released plugin zip |
| `.gitignore` | un-ignore the root lockfile (see below) |

**No production code changed.** `assets/js/twoinc.js` is untouched.

## The harness assembles the real browser

jsdom + the **real** jQuery + the **real** selectWoo widget, with WooCommerce's `wp_localize_script` globals (`wc_country_select_params`, `window.twoinc`) stubbed because they have no npm distribution. The plugin source is then evaluated in global scope exactly as a `<script>` tag would evaluate it.

That is deliberate rather than convenient. The reason the plugin owns a custom `ajax.transport` at all is a property **of** select2's ajax adapter, not of our code: its failure handler treats any jqXHR with `status 0` as a cancellation, and a jQuery `timeout` also reports `status 0`. The "unavailable" message is likewise rendered by the widget's own results adapter via the `results:message` channel. A hand-written mock would have to reproduce both correctly to catch either bug — which is precisely the assumption that let them ship.

selectWoo is not published to npm under any name, so the devDependency is a GitHub tarball pinned to a commit SHA. npm records that with a real integrity hash; a `git+` dependency gets rewritten to `git+ssh://` with the integrity check skipped, which would break `npm ci` on a runner with no SSH key.

## Coverage

- **a timeout is not a cancellation** — `timeout`/`error`/`parsererror` raise the unavailable message, `abort` stays silent. The premise is asserted too: both a timeout and an abort carry `status 0`.
- **`degraded === true`** yields the synthesised empty result set *and* the message; an **absent** field reads as false, and so does every truthy non-`true` value.
- **the `companySearchSeq` guard on all three callbacks** — success, failure and `always`. The success-path guard was added in a review round on TWO-25232; it now has a test.
- **`processResults` does not assume `response.items` is an array** — eleven malformed bodies, none of which throws. A throw there happens inside select2's query pipeline, so it surfaces as a dropdown stuck on "Searching…" rather than as an error.
- the request envelope (the 30s timeout actually reaching `jQuery.ajax`, country, `limit`/`offset`, term decoding) and the full spinner lifecycle.

## The tests were checked against mutations

Green tests prove nothing on their own, so seven mutations of `assets/js/twoinc.js` were each confirmed to fail at least one test: silencing the abort branch, dropping either supersession guard, dropping the `always` guard, weakening `degraded === true` to truthiness, removing the `Array.isArray` guard, and removing the request timeout. The file was restored after each.

## Two things worth a reviewer's eye

1. **`.gitignore` change.** The root `package-lock.json` was ignored by a blanket `package-lock.json` line dating from an unrelated 2024 commit. `npm ci` needs a committed lockfile and `setup-node`'s cache keys off it, so this un-ignores that one path only. `tests/e2e/package-lock.json` is already tracked, force-added past the same line — so committed lockfiles are the existing precedent here.
2. **Two characterisation tests, labelled as such in the code and the README.** Neither is fixed here, because this is a test-only PR: `processResults` throws on an item with no `national_identifier` (the body-level guard does not cover a malformed *hit*), and `client`/`client_v` never reach the company-search query string (`constructTwoincUrl()` sets them as properties, and the caller hands it a `URLSearchParams`, whose *entries* are what get copied). Say the word and either becomes its own ticket.

## Verification

```
$ npm ci --no-audit --no-fund && npm run test:js
PASS tests/js/company-search-results.test.js
PASS tests/js/company-search-transport.test.js

Test Suites: 2 passed, 2 total
Tests:       59 passed, 59 total
Time:        2.14 s
```

`make test-js` and `make help` both verified locally.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -6,6 +6,8 @@ cypress/
 docker/
 docs/
 snippets/
+node_modules/
+tests/js/
 
 # Files to ignore
 .distignore
@@ -17,6 +19,8 @@ snippets/
 bumpver.toml
 dev-requirements.txt
 docker-compose.yaml
+package.json
+package-lock.json
 README.md
 Makefile
 CODEOWNERS
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -32,3 +32,29 @@ jobs:
 
       - name: Run unit tests
         run: php tests/unit/run.php
+
+  jest:
+    name: Jest (Node 20)
+    runs-on: ${{ vars.RUNNER_STANDARD }}
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "20"
+          cache: "npm"
+
+      - name: Install JS devDependencies
+        run: npm ci --no-audit --no-fund
+
+      # Real gate, deliberately not continue-on-error: the company-search
+      # regressions these tests pin (a timeout rendering as "no companies
+      # found", a superseded response overwriting a newer one, a throw inside
+      # processResults leaving the dropdown stuck on "Searching…") were all
+      # invisible to the PHP suite, so an advisory JS job would have caught
+      # none of them.
+      - name: Run Jest
+        run: npm run test:js
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,11 @@ docker/config/
 *.env.json
 .DS_Store
 package-lock.json
+# ...except the root one (TWO-25244). The Jest job runs `npm ci`, which needs a
+# committed lockfile, and `setup-node`'s npm cache keys off it. It also carries
+# the integrity hash pinning the selectWoo tarball. tests/e2e/package-lock.json
+# is tracked too, force-added past the line above.
+!/package-lock.json
 languages/*.php
 screenshots/
 videos/
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 # Copy .env.example to .env first; docker compose reads it natively.
 
 .PHONY: help install configure run debug proxy stop clean logs logs-wpcli \
-	test-unit test format archive bump patch minor major \
+	test-unit test-js test format archive bump patch minor major \
 	e2e-install e2e-test e2e-test-headed phpcs phpstan
 
 .DEFAULT_GOAL := help
@@ -66,6 +66,16 @@ clean:
 ## Run the unit test harness (same suite CI runs)
 test-unit:
 	docker run --rm -v "$(CURDIR)":/app -w /app php:8.2-cli php tests/unit/run.php
+
+## Run the browser-JS test suite (same suite CI runs; needs host Node 20+)
+# Deliberately NOT wired into `test`, mirroring prestashop-plugin: `test` runs
+# in a php container and needs no Node, and the canonical target set predates
+# this suite.
+test-js:
+	@if [ ! -d node_modules ] || [ package-lock.json -nt node_modules ]; then \
+		npm ci --no-audit --no-fund; \
+	fi
+	npm run test:js
 
 ## Run the unit test harness (same suite CI runs)
 test: test-unit
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4505 @@
+{
+  "name": "two-inc/woocommerce",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "two-inc/woocommerce",
+      "devDependencies": {
+        "jest": "^29.7.0",
+        "jest-environment-jsdom": "^29.7.0",
+        "jquery": "^3.7.1",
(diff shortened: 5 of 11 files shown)


dgjlindsay:
## Adversarial self-review

Three rounds over my own diff. Findings and what I did about them.

### Round 1 — four real findings, all fixed

| Finding | Fix |
| -- | -- |
| The abort path was only ever driven through the failure handler directly. In reality select2 cancels a search by calling `abort()` on the jqXHR the transport **returned** — so the returned handle being the cancellable one was assumed, not asserted. | New test drives `run.jqXHR.abort()` and asserts silence, no results, and the spinner down. Also puts the harness's `aborted` flag to work, which was previously set and never read. |
| `jest.config.js` claimed the suite "restores its own spies by hand". It uses no jest mocks at all — it stubs `jQuery.ajax` and restores it in `afterEach`. | Comment corrected to say what `restoreMocks`/`resetMocks` are actually there for: the first test that reaches for `jest.spyOn`. |
| The harness reset `companySearchSeq = 0` with a comment implying that is what keeps it from leaking. It is not — every `loadTwoinc()` re-evaluates the source and yields a fresh helper object. | Comment corrected; the line stays as a guard against a future decision to memoise the eval, which is the only way that counter could become shared state. |
| README did not mention that pinning selectWoo to a GitHub tarball makes the CI job depend on `codeload.github.com` as well as the npm registry. | Stated plainly, with the trade-off. |

`loadTwoinc`'s `@returns` also omitted `dom`; corrected.

### Round 2 — no new findings

Re-read the whole diff with fresh eyes. Specifically checked and found nothing wrong with: the `.gitignore` negation ordering (verified with `git check-ignore -v` — line 15 wins over line 10, and the lockfile is committed), `.distignore` folder-pattern style against the existing entries, the `name` field's `two-inc/...` shape (matches both sibling repos; npm only validates strictly on publish), `permissions`/`timeout-minutes` against the sibling job in the same workflow, and the shared `concurrency` group.

Order-independence checked separately rather than assumed: three runs under `--randomize`, plus each file run in isolation. Stable at 60.

### Round 3 — independent lens: would these tests have caught the real bugs?

Green tests prove nothing by themselves, so the suite was validated twice from the outside.

**Mutation testing.** Seven mutations of `assets/js/twoinc.js`, each confirmed to fail at least one test, file restored after each: silencing the `abort` branch (1 fail), dropping the supersession guard from the success path (2), from the failure path (1), from `always` (1), weakening `degraded === true` to truthiness (3), removing the `Array.isArray` guard (8), removing the request `timeout` (1).

**Against the pre-fix source.** Checked out `assets/js/twoinc.js` as it was immediately before PR #391 / TWO-25232 landed and ran the suite unchanged against it: **44 of 60 fail.** The suite genuinely pins that fix rather than describing whatever the file happens to do now.

No new findings in round 3 either. Calling it clean there.

### Verification quoted, not claimed

Real CI, `Jest (Node 20)` job, 22s:

```
added 334 packages in 2s
PASS tests/js/company-search-transport.test.js
PASS tests/js/company-search-results.test.js
Test Suites: 2 passed, 2 total
Tests:       60 passed, 60 total
```

Also run locally under **Node 20** in a container (`docker run node:20`), not just the Node 24 on this machine, with a clean `npm ci` from the committed lockfile — 60 passed. `pre-commit` and `parity-check` are green, so the prettier gate and the canonical-Makefile-target check both accept the change.

### Left deliberately undone

- **`tests/unit/` and `tests/e2e/` still ship in the released zip.** I excluded `tests/js/` (mine) plus `node_modules/` and the manifests. Excluding the other two is a pre-existing gap and would change the release artifact for files this PR did not add, so it is reported rather than done.
- **`make test-js` is not wired into `make test`**, mirroring `prestashop-plugin`: `make test` runs the PHP suite in a container and needs no Node on the host.
- **Two characterisation tests** are labelled as such in the code and the README. `processResults` throws on a hit with no `national_identifier`, and `client`/`client_v` never reach the company-search query string. Neither is fixed here — this is a test-only PR — and both are pinned so that fixing them is a deliberate change with a test that flips.

_Review by Claude._

PR #393, opened 2026-07-29
TWO-25249/fix: guard company-search rendering and send client attribution

dgjlindsay:
Closes TWO-25249. Fixes the two defects the Jest suite (#392) pinned as characterisation tests and deliberately left unfixed.

## 1. `processResults` threw on a hit with no `national_identifier`

`item.national_identifier.id` was read unguarded. That field is `CompanyIdentifier | None` in the search response schema, so a company with no identifier in its home registry is a normal hit, not a malformed one — and the `id` inside it can be null or empty. The body-level `Array.isArray` guard above only covers a malformed *body*.

The throw happened inside select2's query pipeline, which is the expensive part: it took the **whole result list** down and left the dropdown on "Searching…" rather than surfacing an error.

Such a hit now renders as the company name / highlight alone, with `company_id: ""`. Rationale: the identifier is what the buyer uses to tell two similarly-named companies apart, so losing it is a real cost — but skipping the hit removes a company the buyer may need to select, which is worse. `company_id` remains a required checkout field, so the buyer types the organisation number themselves; `.val("")` is what jQuery already did with the old `undefined`.

## 2. `client` / `client_v` never reached the company-search query string

`constructTwoincUrl()` set them as plain properties on the object it was handed. The search's `url()` callback hands it a `URLSearchParams`, and `new URLSearchParams(params)` copies entries, not JS properties — so both were silently dropped for that one caller. Every plain-object caller did send them.

Now normalised to a `URLSearchParams` and written through `set()`, which handles both shapes and mutates neither.

Query string, not a header, on purpose: a custom header makes the request non-simple and buys a CORS preflight on **every keystroke**. And this is the only platform attribution available for this endpoint — the widget is client-side, so the user-agent is the shopper's browser.

**Are the extra params safe?** Yes. `GET /companies/v2/company` binds its query into a pydantic model that does not set `extra="forbid"`, so unmodelled query params are ignored rather than rejected. No 400 risk.

## Tests

`Jest (Node 20)`, 65 tests, all green. The two characterisation tests now assert the corrected behaviour, plus: the null and empty-`id` variants, that one unusable hit does not cost the buyer the other results in the same response, and that the attribution params are *added to* the search params rather than replacing them. The characterisation framing is gone from `tests/js/README.md`.

Mutation-checked — reverting each fix independently turns the corresponding tests red:

```
revert the national_identifier guard  -> 5 failed, 60 passed
  a hit with no usable national_identifier > national_identifier absent / null / id null / id empty
  a hit with no usable national_identifier > does not take the rest of the result list down with it

revert constructTwoincUrl             -> 2 failed, 63 passed
  request url > identifies the plugin and its version to the API
  request url > keeps the search params it was given alongside them
```

## Not in scope

- The same unguarded `national_identifier.id` read exists in `magento-plugin` (`address-autocomplete.js`, `gateway_method.js`) and needs its own ticket.
- `/autofill/v1/buyer/current` sends no client attribution either, but via a raw `fetch()` that never went through `constructTwoincUrl()` — a separate omission.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -14,13 +14,25 @@ let twoincUtilHelper = {
   },
 
   /**
-   * Construct url to Twoinc checkout api
+   * Construct url to Twoinc checkout api.
+   *
+   * `client` / `client_v` identify this plugin and its version to the API, and
+   * are the only attribution the company-search endpoint can get: the widget
+   * runs in the buyer's browser, so the user-agent is the shopper's. They go
+   * in the query string rather than a header on purpose — a custom header
+   * makes the request non-simple and buys a CORS preflight per keystroke.
+   *
+   * `params` may be a plain object or a URLSearchParams. It used to be
+   * assigned to as an object either way, which silently dropped both fields
+   * for every URLSearchParams caller (the company search): `new
+   * URLSearchParams(existing)` copies entries, not JS properties. Normalising
+   * first, and going through set(), covers both shapes and mutates neither.
    */
   constructTwoincUrl: function (path, params) {
-    if (!params) params = {};
-    params["client"] = window.twoinc.client_name;
-    params["client_v"] = window.twoinc.client_version;
-    return window.twoinc.twoinc_checkout_host + path + "?" + new URLSearchParams(params).toString();
+    const searchParams = new URLSearchParams(params || {});
+    searchParams.set("client", window.twoinc.client_name);
+    searchParams.set("client_v", window.twoinc.client_version);
+    return window.twoinc.twoinc_checkout_host + path + "?" + searchParams.toString();
   },
 
   /**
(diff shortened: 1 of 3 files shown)


dgjlindsay:
## Adversarial self-review

One round, one finding, refuted below — nothing actionable.

**🟡 `identifier` guard treats a falsy-but-valid `id` (`0`, `false`) as absent.** Refuted by the response contract: `national_identifier.id` is a non-empty **string** in the search schema, and it is JSON-serialised through a field serializer that prefixes it by identifier type. There is no path on which it arrives as `0` or `false`, and the string `"0"` is truthy. Guarding on truthiness is therefore equivalent to guarding on `!= null && !== ""` here, and the `String()` wrapper keeps a numeric value rendering as before if the contract ever loosened.

Checked and clean:

- **the other three `constructTwoincUrl()` call sites** (`/v1/order_intent`, `/companies/v2/company/{lookup_id}`, `/v1/payment_terms`) are byte-identical before and after — all three pass a plain object or nothing, and both the old and new code route that through `new URLSearchParams(obj)`. Verified rather than reasoned: `new URLSearchParams({x: undefined, y: 1}).toString()` gives `x=undefined&y=1` on both paths, so even the `merchant_id`/`merchant_short_name` optional-chaining case is unchanged.
- **param order and duplicate keys.** `client`/`client_v` still land last, and `set()` overwrites rather than appends, matching the old property assignment.
- **the caller's object is no longer mutated.** The old code wrote `client`/`client_v` onto whatever it was handed; the new code copies first. No caller depended on the mutation.
- **`company_id: ""` downstream.** Every consumer already treated the old `undefined` the same way `""` is treated: `insertFloatingCompany()` early-returns on a falsy id, `isReadyApprovalCheck()` early-returns on a falsy `organization_number` so no order intent is created for a half-identified company, and `$companyId.val("")` is exactly what jQuery already did with `undefined`. The only behaviour change is the absence of the crash.
- **`id: item.name` collisions** between two hits with the same name are pre-existing and untouched by this diff.
- **test quality.** Both new assertions fail on a revert (5 and 2 tests respectively, output in the PR body), so neither is a tautology. The `lookup_id: undefined` in the `toEqual` is the real shape of a hit built without one, not a brittle stand-in.

Not fixed here, needs its own ticket: the same unguarded `national_identifier.id` read exists in `magento-plugin` (`address-autocomplete.js:131`, `gateway_method.js`).

_by Claude_

PR #395, opened 2026-07-29
fix(admin): move surcharge-grid help text below the grid, match Magento copy (ABN-476)

dgjlindsay:
## Why

Two defects on the Payment Surcharge Configuration field, both parity gaps against the Magento admin:

1. The help text rendered **above** the grid. Magento renders it **below**.
2. The wording had no Magento counterpart, so the same setting was explained differently on the two platforms (and the Dutch translation diverged too).

## What

`generate_two_surcharge_grid_html()` now renders, below the table and in this order, the copy from `magento-plugin`'s `view/adminhtml/templates/system/config/field/surcharge-grid.phtml`:

| String | Source |
|---|---|
| `Amounts are shown in %s.` | `surcharge-grid.phtml` `.surcharge-grid__currency-note` |
| `Enter the amount you want to charge your customer. Max %s.` | `.surcharge-grid__helper-text--fixed` |
| `Enter the percentage of the fee you want to charge your customer. Max: %s.` | `.surcharge-grid__helper-text--percentage` |
| `Enter the amount and percentage of the fee you want to charge your customer. Max %1$s / %2$s.` | `.surcharge-grid__helper-text--fixed_and_percentage` |
| `Enter a limit amount (in %s) if you do not want to charge your customer more than a specific amount. Leave the limit field empty if you don't want to impose a limit amount.` | `$limitSentence`, appended to both percentage-bearing variants |

Magento's placeholders are `%1`/`%2`; adapted to WordPress `%s` / `%1$s`.

As in Magento, all three method-keyed paragraphs are rendered hidden and `assets/js/admin.js` shows the one matching the selected surcharge method — mirroring `view/adminhtml/web/js/surcharge-grid.js`.

## Reviewer notes

- **The generic sentence is gone.** `Per payment term: a fixed amount, a percentage of the order, and an optional cap on the percentage portion. Leave a cell blank for zero. Fixed amounts and caps are in the store currency and are not converted for multi-currency stores.` was removed from the field definition, because Magento has no equivalent and the new currency note carries the currency statement. That does drop the explicit "not converted for multi-currency stores" caveat, which is genuinely WooCommerce-specific (Magento converts). Flagging it: say so if it should be kept as an extra line.
- A `description` set by a `wc_two_form_fields` filter is still rendered — now below the table, ahead of the parity notes. The seam is intact.
- **Fixed maximum keeps its currency guard.** Claimed only when the cap's currency matches the store currency, matching what save-validation actually enforces (Woo does no FX conversion). On a mismatch the fixed-only paragraph is not rendered at all and the combined variant degrades to the percentage-only wording, rather than claiming a maximum that will not be enforced. Magento renders a bare `Max .` in the equivalent no-limit case; that was deliberately not ported.
- **Percentage ceiling** is `100%`, the bound `validate_two_surcharge_grid_field()` enforces, so the claimed and enforced maxima cannot drift.
- `nb_NO` / `sv_SE` untouched — those catalogues never carried any of the surcharge-grid strings (126 entries vs `nl_NL`'s 190), so adding five here would have been arbitrary.

## Test

`testSurchargeGridHelpTextOmitsMaxOnCurrencyMismatch` tightened: it previously asserted the substring `Max` was absent on a currency mismatch, which the always-enforceable `Max: 100%` now trips. It now asserts no *fixed* maximum is claimed (`Max EUR` / `Max NOK` absent, fixed-only paragraph absent) and that the percentage ceiling is present.

```
$ php tests/unit/run.php
...
All tests passed.
$ pre-commit run --files class/WC_Twoinc.php assets/js/admin.js tests/unit/run.php languages/*.po languages/*.pot
prettier.................................................................Passed
```


dgjlindsay:

**Round 1** — one finding, refuted:

- *"`updateGridColumns()` is never called on initial page load, so the help paragraphs start hidden until the merchant touches the surcharge-method select."* — **Not so.** The init block calls `updateGridRows()`, whose last statement is `updateGridColumns()`. Its only early return is `$grid.length === 0`, and `updateGridColumns()` early-returns on the same condition, so there is no page state where one runs without the other. No change made.

**Round 2** — no issues. Specifically checked, all clean:

- `esc_html()` on the composed sentences, which carry a decoded `€` and the apostrophe in "don't" — no double-encoding or mangling;
- the literal `%` in `'100%'` passing through `sprintf`'s `%s` — no format hazard;
- `/* translators: */` comments still attached to the `__()` calls they document after the refactor into `$help_text`;
- `.pot` / `.po` php-format flags, and no dangling reference left by the removed msgid;
- the `twoinc-surcharge-grid-help--<method>` class ↔ `admin.js` selector coupling across all four method slugs;
- nothing else read the `surcharge_grid` field's `description` key.

Verified locally at the CI's own tool version rather than by iterating through the pipeline:

```
$ php phpcs.phar   # PHPCS 4.0.1, the version static-analysis.yaml pins
phpcs exit=0
$ php tests/unit/run.php
All tests passed.
```

The first push failed `phpcs` on two indentation errors from a nested ternary built inline in the markup; the follow-up commit composes the three sentences above `ob_start()` instead, which is both clean and more readable. No behaviour change in that commit.

PR #396, opened 2026-07-29
fix(checkout): chip currency symbol and singular/plural term labels (ABN-468)

dgjlindsay:
## Why

Two parity gaps in the buyer-facing payment-terms chips, against Magento's Luma renderer (`view/frontend/web/js/view/payment/method-renderer/gateway_method.js` + `view/frontend/web/template/payment/gateway_method.html`) — the same template the Amasty and Fire checkouts render.

## 1. Currency symbol, not code

The chip fee label was `"+" + fee.buyer_fee_share + " " + fee.currency` → **"+12.50 EUR"**. Magento renders `'+' + priceUtils.formatPrice(amount, quote.getPriceFormat())` → **"+€12.50"**: the symbol, in the store's configured position, with the store's separators.

The fees endpoint now returns `buyer_fee_share_display`, produced by `WC_Twoinc_Payment_Terms::format_fee_amount()`. That wraps `wc_price()` — WooCommerce's single source of the store price format — and strips the markup/decodes the entities, since the chip is rendered with jQuery `text`. Deliberately not reimplemented from `get_woocommerce_currency_symbol()` plus the two separator options: that would be a second copy of the store's price format, free to drift.

The raw amount + currency code remains as the degraded fallback in the JS.

## 2. Singular / plural chip copy

Neither string existed on WooCommerce. Both come from the Luma renderer:

| Case | Behaviour | String | nl_NL |
|---|---|---|---|
| exactly 1 offered term | label injected **inside** the chip, **no** heading | `Payment Terms %s days` (Magento `Payment Terms %1 days`) | `Betaaltermijn %s dagen` |
| more than 1 | heading immediately **above** the chips | `Selected payment terms` | `Gewenste betaaltermijn` |

Both nl_NL strings are copied verbatim from `magento-plugin`'s `i18n/nl_NL.csv`; only the placeholder syntax is adapted (`%1` → `%s`).

The heading is a `<label class="twoinc-term-chips-heading">` sibling **ahead of** the `role="radiogroup"` container, matching Magento's `<label class="label">` sitting outside `.two-term-chips__container` — not a pseudo-chip inside the group.

### Behaviour change worth a look

The single-term case previously rendered **no chip at all**: the bootstrap gated the whole selector on `is_selector_visible()` (`>1` term), so a shop offering one term showed nothing and applied it silently. Magento shows that term (and its surcharge) as a disabled chip. The bootstrap flag is now `is_enabled()` (`>=1` term).

Whether the buyer can *choose* is deliberately not sent to the client: `render()` derives it from the term list it is handed, which the fees response can revise after the bootstrap was written, so a flag fixed at render time would go stale.

Two consequences worth a look:

- single-term shops now also fire the per-term fee quote, which is what makes the surcharge visible on that chip. `ajax_term_fees` already guarded on `is_enabled()`, and a posted term is validated with `in_array` against the offered set (`WC_Twoinc_Payment_Terms.php:726`, `WC_Twoinc.php:2785`), so nothing new is trusted;
- **`is_selector_visible()` now has no production caller** — it was only ever called from this bootstrap, and is referenced only by its own unit test. Left in place rather than deleted: it is a public static on a class merchants' snippets can reach, and the predicate is still the honest name for "buyer can choose". Say if you would rather it go.

### Note on the string-typing class of bug

The recent finding that Magento's flags interpolate into templates as strings (`'0'` being truthy) has no analogue here. `payment_terms` is a **nested** array in the `wp_localize_script` payload, and `WP_Scripts::localize()` only string-casts top-level scalars — so it is JSON-encoded with types intact. The singular/plural branch is `terms.length`, a real number, in both the PHP bootstrap and the JS.

## Test

New `tests/js/term-chips.test.js` (6 tests) pins the singular label, the absent-heading-when-single rule, the heading text and its position outside the radiogroup, the retraction of a stale heading when the term set shrinks to one, and the symbol fee label plus its code fallback. `wc-harness.js` now exposes `twoincTermChips` (it was already in the eval's export list).

New `testChipFeeAmountCarriesCurrencySymbolNotCode` in `tests/unit/run.php` covers the server-side formatting; `tests/unit/bootstrap.php` gains `wc_price()` / `wp_strip_all_tags()` stubs, with the price stub deliberately not built on the existing `get_woocommerce_currency_symbol()` stub (which returns the code — the very thing under test).

```
$ npm run test:js
Test Suites: 3 passed, 3 total
Tests:       71 passed, 71 total
$ php tests/unit/run.php
All tests passed.
$ pre-commit run --files <changed>
prettier.................................................................Passed
```

`tests/js/README.md`'s "Known gaps" listed term chips as uncovered; corrected to name what is still uncovered (`refresh()`'s fetch half).


dgjlindsay:

**Round 1** — two findings, both fixed in `c8cd421`:

1. **`selectable` was dead.** The bootstrap emitted it and nothing read it. Removed rather than wired up: `render()` derives the single/plural state from the term list it is handed, which the fees response can revise after the bootstrap was written, so a flag fixed at render time would go stale.
2. **Single-term shops now fire the fees ajax and render a chip** where before they rendered nothing. Traced and found safe — `ajax_term_fees` already guarded on `is_enabled()`, and a posted term is validated with `in_array` against the offered set (`WC_Twoinc_Payment_Terms.php:726`, `WC_Twoinc.php:2785`), so nothing new is trusted. But `get_pay_box_description()`'s comment still said the server-rendered hidden input "covers the single-term case, where the chip chooser never renders and JS posts nothing" — no longer true. Rewritten to say what the input is actually for.

**Round 2** — no issues. Specifically checked, all clean:

- **Merge integrity.** #395 landed on `staging` between the two rounds, touching the same `class/WC_Twoinc.php` and the same `nl_NL` catalogue. Both sides survive intact: the ABN-476 `$help_text` / `twoinc-surcharge-grid-help--<method>` block and this branch's `twoinc-term-chips-heading` label. The only conflict was the binary `.mo`, resolved by regenerating with `msgfmt` from the cleanly auto-merged `.po` — which now carries both string sets, no duplicate msgids, `php-format` flags correct.
- `format_fee_amount()` against `wc_price()`'s `&nbsp;` thousand separator (U+00A0 after decoding), negative and zero amounts, and safety in a `wc-ajax` context.
- `tests/js/term-chips.test.js`'s hand-built DOM compared literally against what `class/WC_Twoinc.php` renders — class names and nesting match.
- `role="radiogroup"` around a single disabled button, and the `<label>` without a `for` — noise, not a defect.
- No Playwright selector in `tests/e2e/` depends on the chip fee text or the element order the new `<label>` changes.
- `WC_Twoinc_Payment_Terms::is_selector_visible()` is confirmed to have no production caller now (repo-wide grep incl. `views/`, `snippets/`, `assets/`). Left in place — noted in the PR description as something to say if you'd rather it go.

```
$ npm run test:js
Test Suites: 3 passed, 3 total
Tests:       71 passed, 71 total
$ php tests/unit/run.php
All tests passed.
$ php phpcs.phar    # PHPCS 4.0.1, the version static-analysis.yaml pins
phpcs=0
```

PR #397, opened 2026-07-29
fix(checkout): order optional fields invoice/PO/project/department (TWO-25263)

dgjlindsay:
## What

The optional buyer fields were listed and rendered in reverse. Canonical
cross-plugin order is **invoice email, purchase order number, project,
department**, matching the PrestaShop plugin.

- **Admin pane** — reordered the four entries in `init_form_fields()`.
  WooCommerce renders `form_fields` in array order.
- **Checkout** — swapped the `priority` offsets and reordered the source
  blocks to read the same way.
- **Toggle labels** — adopted the PrestaShop plugin's `Show xxx field`
  wording, replacing `Add input field for "xxx"`.
- **Translations** — `nl_NL`, `nb_NO`, `sv_SE` (`.po` + recompiled `.mo`)
  and the `.pot`.

The order note is untouched. WooCommerce core owns it as `order_comments`;
the plugin adds neither a field nor a toggle for it.

## Rendered-order evidence

Not source order — captured from a real WordPress 6.8 + WooCommerce 9.9.5
install by rendering the actual settings table through
`WC_Settings_API::generate_settings_html()` and by reading
`WC_Checkout::get_checkout_fields('billing')` after its priority sort.

Admin pane, rows 11-14:

| | before (`staging`) | after |
|---|---|---|
| 11 | Add input field for "Department" | Show Invoice email field |
| 12 | Add input field for "Project" | Show Purchase order number field |
| 13 | Add input field for "Purchase order number" | Show Project field |
| 14 | Add input field for "Invoice email address" | Show Department field |

Checkout billing fields, rows 7-10:

| | before (`staging`) | after |
|---|---|---|
| 7 | `department` (prio 32) | `invoice_email` (prio 32) |
| 8 | `project` (prio 33) | `purchase_order_number` (prio 33) |
| 9 | `purchase_order_number` (prio 34) | `project` (prio 34) |
| 10 | `invoice_email` (prio 35) | `department` (prio 35) |

## Checks

`make test-unit` (all pass), `make phpcs` (0 errors), `make phpstan` (no errors).

Linear: TWO-25263


dgjlindsay:
## Self-review summary

Four adversarial review rounds. Round 4 found one LOW and nothing else; rounds 1-3 findings are all addressed or consciously declined below.

### Fixed during review

- **Toggle wording contradicted the field it gates.** The toggles named the purchase-order field "innkjøpsordrenummer" / "inkoopordernummer" / "inköpsordernummer" while the checkout label is `PO-nummer` in all three locales — a merchant enabled a toggle naming one thing and got a field naming another. Each locale now uses its own field label's terminology, with one construction across all four toggles.
- `invoice_email`'s guard used `==` where its three siblings use `===`. Harmless, but they are now one contiguous block so the odd one out reads as an oversight.
- Catalogue entries were left in their old alphabetical slot; moved beside `Show "What is %s" link in checkout` so the next `make-pot` does not relocate them as churn.
- Doubled preposition in the `nb_NO` / `sv_SE` invoice-email strings.
- `assets/js/twoinc.js` and the user-profile pane still listed the old order. Both order-insensitive at runtime, but the new comments call this order canonical and the listings should not contradict them.
- Two `docker/config` templates reordered to match `README.md`.

### Rendered-order evidence

Source order and priority values are not evidence, so this was captured from a real WordPress 6.8 + WooCommerce 9.9.5 install — the admin table rendered through `WC_Settings_API::generate_settings_html()`, and the checkout fields read from `WC_Checkout::get_checkout_fields('billing')` after its priority sort.

Admin pane rows 11-14:

| | before (`staging`) | after |
|---|---|---|
| 11 | Add input field for "Department" | Show Invoice email field |
| 12 | Add input field for "Project" | Show Purchase order number field |
| 13 | Add input field for "Purchase order number" | Show Project field |
| 14 | Add input field for "Invoice email address" | Show Department field |

Checkout billing rows 7-10:

| | before (`staging`) | after |
|---|---|---|
| 7 | `department` (prio 32) | `invoice_email` (prio 32) |
| 8 | `project` (prio 33) | `purchase_order_number` (prio 33) |
| 9 | `purchase_order_number` (prio 34) | `project` (prio 34) |
| 10 | `invoice_email` (prio 35) | `department` (prio 35) |

### Catalogue integrity

Rebased twice while `staging` moved underneath. The `.po` text-merges but the binary `.mo` cannot, and resolving that toward this branch would silently drop the other side's strings — so all three `.mo` were recompiled from the merged `.po` each time. Verified after the final rebase: `msgfmt --check` clean on all three; each `.mo` byte-identical in content to its `.po`; and a msgid **set** diff against `staging`'s `.mo` shows exactly the four renamed labels removed and their four replacements added, nothing else lost. `msgcmp` gap counts are unchanged from `staging`, so this branch does not make catalogue completeness worse.

### Declined, deliberately

- **`invoice_email` keeps `form-row-wide` without `hidden`**, unlike its three siblings, so there is a brief flash before `twoinc.js` takes over — more noticeable now the field is first in the group. Adding `hidden` is a behaviour change beyond this PR's scope and carries a real downside: if the JS fails to run, the field becomes permanently invisible where today it shows. Left fail-open.
- **`.pot` source-line refs.** The four moved entries' refs drift on every rebase, and the rest of the file is uniformly stale from a much older extraction (`Pre-approve buyer during checkout` points at line 1664, actually 3411). Refs are not compiled into `.mo` and have no runtime effect. Hand-maintaining four of them in a file where ~200 are wrong buys nothing; the real fix is a `wp i18n make-pot` regeneration, which this repo has no target for. An earlier commit message in this branch claims these were "corrected" — that was true when written and is no longer, after the rebases.
- **Pre-existing:** `tracking_id` sits at priority 20, tying WooCommerce core's `billing_last_name`. Hidden field, no visible symptom, unspecified tie-break. Outside this change.

### Checks

`make test-unit`, `make test-js` (71), `make phpcs` (0 errors), `make phpstan` (no errors) — all green on the final rebased commit.

PR #398, opened 2026-07-29
fix(admin): wrap surcharge grid help text to the grid width

dgjlindsay:
ABN-476 follow-up.

## Problem

The per-term surcharge grid's help text and currency note sit below the grid (as of PR #395), but they were laid out against the full `<td class="forminp">` of the WooCommerce settings table. That cell runs to the right page margin, so the paragraphs did too, while the grid itself stopped at 620px.

## Fix

Grid, notes and help paragraphs now share **one** width container, `.twoinc-surcharge-grid-container`. The `max-width` lives on the container (`assets/css/admin.css`) and the grid table is `width: 100%` of it, so the container is the single width source and the two cannot drift apart.

This is the mechanism the Magento gateway already uses: `#surcharge-grid-container` in `view/adminhtml/templates/system/config/field/surcharge-grid.phtml`, with `.surcharge-grid { width: 100% }` in `_surcharge-grid.less` and the `.note` paragraphs as siblings inside the same box. Copied rather than reinvented.

## Verification

Headless Chromium render of the real `generate_two_surcharge_grid_html()` output plus the real `admin.css`, inside a `.form-table` harness, 1440px viewport:

| element | before | after |
|---|---|---|
| settings cell (`td.forminp`) | 1050px, right edge 1260 | unchanged |
| grid table | 620px, right edge 840 | 620px, right edge 840 |
| help paragraph | 1030px, right edge 1250 | **620px, right edge 840** |
| currency note | 1030px, right edge 1250 | **620px, right edge 840** |

The help block's right edge now lands exactly on the grid's. Note this is a local harness approximating wp-admin's `.form-table` rules, not the live staging admin — the branch is not deployed there.

New unit test `testSurchargeGridNotesShareTheGridsWidthContainer` asserts the container wraps the grid and every note/help paragraph, and that the table carries no width of its own. Full `php tests/unit/run.php` suite passes.

### DIFF
--- a/assets/css/admin.css
+++ b/assets/css/admin.css
@@ -40,3 +40,20 @@ a.setting-dropdown {
     max-width: 300px;
   }
 }
+
+/*
+ * Surcharge grid: the container is the SINGLE width source for the grid
+ * and for the notes/help paragraphs that sit below it. The table is
+ * width:100% of the container and the paragraphs are its children, so
+ * the help text wraps at exactly the grid's width instead of running out
+ * to the page margin of the wide <td class="forminp">. Mirrors Magento's
+ * #surcharge-grid-container (.surcharge-grid is width:100%, .note
+ * paragraphs are siblings in the same box). Change the width here only.
+ */
+.twoinc-surcharge-grid-container {
+  max-width: 620px;
+}
+
+.twoinc-surcharge-grid-container .twoinc-surcharge-grid {
+  width: 100%;
+}
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -1245,8 +1245,19 @@ if (!class_exists('WC_Twoinc')) {
             <tr valign="top" class="twoinc-surcharge-grid-field">
                 <th scope="row" class="titledesc"><label><?php echo wp_kses_post($data['title']); ?></label></th>
                 <td class="forminp">
+                    <?php // Grid, notes and help text share ONE width source: the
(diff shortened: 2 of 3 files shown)


dgjlindsay:
## Adversarial self-review — round 1, clean

Reviewed the full diff for markup nesting, JS coupling, CSS specificity and escaping. No findings.

Two items the reviewer flagged as unverifiable from the diff alone, both since checked against the branch:

- **`admin.css` is enqueued on this page.** `WC_Twoinc::twoinc_admin_styles_scripts()` does `wp_enqueue_style('twoinc.admin', …/assets/css/admin.css)` and is hooked on `admin_enqueue_scripts`. The same stylesheet already carries the `.woocommerce_page_wc-settings .forminp` rules that visibly apply today, so the new rules land on the settings screen.
- **No JS reads the grid's width.** `assets/js/admin.js` touches the grid only via `.toggle()`, `data-field-key` and row/column class selectors — no `outerWidth`/`innerWidth`/`.css()` width read, and no reference to `620`. Removing the inline `max-width` breaks nothing there. The conditional inline `display:none` is preserved, so `$grid.toggle()` behaves as before.

Also confirmed: no doc in the repo describes the grid's help-text placement or width, so nothing to update alongside.

Measured evidence (headless Chromium, 1440px viewport, real render output + real stylesheet) is in the PR description: help paragraph 1030px → 620px, right edge 1250 → 840, exactly matching the grid.

PR #399, opened 2026-07-29
refactor(checkout): drop the unused term-selector visibility helper (ABN-468)

dgjlindsay:
ABN-468 follow-up.

`WC_Twoinc_Payment_Terms::is_selector_visible()` has had no production caller since the chip bootstrap stopped sending a selectable flag — `render()` derives whether the buyer can choose from the term list it is handed, because the fees response can revise that list after the bootstrap is written. The only reference left was its own unit test.

- remove the method
- remove `testPaymentTermsSelectorVisibleOnlyWithMultiple` and its registration; the `is_enabled()` assertions it also carried are already covered by the existing terms test, so no coverage is lost
- de-stale the refresh comment in `prepare_twoinc_object()`, which named the removed helper as an example cache-only seam reader

No behaviour change. `php tests/unit/run.php` passes.

Casing check (same ticket): the shipped chip heading is `Selected payment terms`, lower-case after the first word, matching the Luma renderer. No change needed.

### DIFF
--- a/class/WC_Twoinc_Checkout.php
+++ b/class/WC_Twoinc_Checkout.php
@@ -235,8 +235,8 @@ if (!class_exists('WC_Twoinc_Checkout')) {
 
             // Checkout render is the sanctioned refresh point for the
             // backend term list (TWO-24812) — refresh once here; the
-            // cache-only seam reads below (is_selector_visible,
-            // get_selected_term, …) then see the fresh list.
+            // cache-only seam reads below (is_enabled, get_selected_term,
+            // …) then see the fresh list.
             $offered_terms = WC_Twoinc_Payment_Terms::get_available_terms($this->wc_twoinc, true);
 
             // TODO: Make this dynamic based on active merchant payee accounts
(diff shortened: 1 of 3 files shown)


dgjlindsay:
Self-review (repo does not enforce a separate reviewer; `reviewDecision` is empty).

Adversarial round on `origin/staging...HEAD`, no findings:

- `is_selector_visible` / `selectorVisible`: zero hits repo-wide after the deletion, and zero in the overlay repo that consumes this plugin — nothing external was calling it.
- Coverage not lost: the deleted test also asserted `is_enabled()`, but every case it covered (including the single-term case) is already asserted in the existing backend-intersect terms test.
- Test registry list updated in step with the removed method, so the runner does not reference a missing callable.
- The comment edit in `prepare_twoinc_object()` is accurate: `is_enabled` is a real cache-only reader of that seam.
- `php tests/unit/run.php` green locally; all CI checks pass (phpcs, phpstan, pre-commit, unit-tests 7.4-8.4, Jest, e2e, boot/upgrade smoke).

Behaviour unchanged — pure dead-code removal.

Casing check for the same ticket: the shipped chip heading is `Selected payment terms`, i.e. sentence case, matching the Luma renderer. Nothing to change.

<sub>by Claude</sub>

PR #400, opened 2026-07-29
TWO-25268/fix: surcharge grid states the STORE currency, and the cap is gated on it

dgjlindsay:
## Why

The surcharge grid's amounts are denominated in the **store** currency — the saved `woocommerce_currency` option, which is what `WC_Twoinc_Payment_Terms::build_buyer_fee_share()` converts *from* via the Two FX layer.

Two places read the **active** currency of the request (`get_woocommerce_currency()`) instead. Under a multicurrency plugin, an admin session in a non-default currency therefore saw:

1. "Amounts are shown in USD" above a grid the plugin reads as EUR — the one sentence whose entire job is stating the denomination, wrong.
2. The funding-partner cap silently not enforced, and its "Max EUR 25" sentence gone.

## Two commits, deliberately separate

**1 — display only, no behaviour change.** The currency note and the symbol in the limit sentence now derive from `get_option('woocommerce_currency')`. The cap comparison is explicitly pinned to the active currency in this commit so it stays byte-identical in behaviour.

**2 — the behaviour-bearing change.** The cap comparison moves to the store currency in both the grid render and the save-validation. This is a **pre-existing** bug, not one introduced by commit 1, which is why it is split out.

The store currency is the right answer because the amounts the cap bounds are the ones typed into the grid, and those are store-denominated. The in-code comment on both sites *already* said "when the cap's currency differs from the **store** currency" — the code contradicted its own documented intent. The old behaviour was fail-open on the value that actually gets saved.

A genuine store-currency mismatch still skips the cap and defers to the backend. Unchanged.

## Tests

- New: the note and the limit-sentence symbol name the store currency while the active currency diverges.
- Updated: the two existing mismatch tests asserted the mismatch by moving the *active* currency while calling it the store currency. They now move the store currency, and each gains an active-currency-only case asserting the cap survives it.
- Both new assertions were confirmed to fail against pre-fix code, not just pass against the fix.

`make test-unit` green, `make phpcs` exit 0.

## Scope

Admin-facing. No buyer-visible change. No migration, no config key, no version bump.

Fail-closed FX harmonisation (TWO-25269) is deliberately a separate PR so this one-line-class fix is not held behind its design work.

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -1200,8 +1200,7 @@ if (!class_exists('WC_Twoinc')) {
             // told the grid is denominated in a currency the plugin never
             // reads it as.
             $currency_code = strtoupper((string) get_option('woocommerce_currency'));
-            $active_currency_code = strtoupper((string) get_woocommerce_currency());
-            $fixed_limit_label = $fixed_limit && $fixed_limit['currency'] === $active_currency_code
+            $fixed_limit_label = $fixed_limit && $fixed_limit['currency'] === $currency_code
                 ? $this->format_surcharge_limit_label($fixed_limit)
                 : '';
             // Percentage ceiling. Mirrors Magento's
@@ -1315,9 +1314,14 @@ if (!class_exists('WC_Twoinc')) {
             // Only enforceable when the cap's currency matches the store
             // currency — Woo does no FX conversion (unlike Magento), so on
             // a mismatch the cap is skipped here and the backend enforces.
+            // The posted amounts are STORE-denominated, so the comparison is
+            // against the saved woocommerce_currency option and NOT against
+            // get_woocommerce_currency() (the active request currency), which
+            // would let an admin session in a non-default currency skip the
+            // cap entirely (TWO-25268).
             $max_fixed = null;
             $limit = $this->get_merchant_surcharge_limit(true);
-            if ($limit && $limit['currency'] === strtoupper((string) get_woocommerce_currency())) {
+            if ($limit && $limit['currency'] === strtoupper((string) get_option('woocommerce_currency'))) {
                 $max_fixed = (float) $limit['amount'];
             }
 
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -1191,8 +1191,17 @@ if (!class_exists('WC_Twoinc')) {
             // backend enforces instead. Omitted entirely when no cap
             // exists or on a currency mismatch.
             $fixed_limit = $this->get_merchant_surcharge_limit(true);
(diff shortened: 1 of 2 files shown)


dgjlindsay:
## Adversarial self-review — 2 rounds, round 2 clean

### Round 1 — 4 findings

**1. Cap enforcement now fires where it previously failed open — ACCEPTED, not changed.**
A store whose active currency differed from the cap's could previously save an over-cap fixed fee; it now gets a validation error naming the cap and the term. That is the point of commit 2. Worth being explicit that the *throw path itself is pre-existing* — the cap already rejects over-cap values for same-currency stores today, so this widens *when* the exception fires, not *what happens* when it does. No new class of behaviour, and the previously-savable value was invalid anyway.

**2-4. Test globals leaked on assertion failure — FIXED at the root (commit 3).**
`bootstrap.php` stubs the STORE currency and the ACTIVE currency from two independent globals, but the harness `reset()` only unset the active one — the store global was never in it, because until this branch no test moved it. The whole suite runs in one process, so a test that set the store currency and then failed an assertion before its own unset leaked a non-default store currency into every subsequent test, turning one real failure into a cascade of unrelated ones.

Fixed by adding the unset to `reset()` rather than wrapping each call site in try/finally — it fixes the leak class at the root and also covers the pre-existing tests that already moved this global.

### Round 2 — no findings

Verified specifically:
- `reset()` is called before **every** test inside the runner's foreach loop, not merely at suite start — so commit 3 does what its message claims.
- Every one of the five tests that sets `__twoinc_test_store_currency` sets it *after* `reset()` within its own body, so none relied on the global persisting. `unset()` on an absent key is idempotent, and the bootstrap default (`EUR`) is the intended fallback.
- No test weakened or silently no-op'd by the change.

Round 2 found nothing new and was not merely cleaning up after round 1's fix, so this is a genuine clean round rather than an oscillating loop.

### Verification

- `make test-unit` green; `make phpcs` exit 0.
- Both new assertions were confirmed to **fail** against pre-fix code, not just pass against the fix — the store-currency mismatch case throws on `origin/staging` exactly where the test expects it to pass.
- Change confirmed complete in scope: the only remaining `get_woocommerce_currency()` reads in `WC_Twoinc.php` are the availability gate's basket currency, which correctly *is* the active currency.

### One judgment call for a human

Finding 1 is the only thing here anyone might want to overrule: a legacy multicurrency store carrying an over-cap grid will have to fix the grid before it can save the payment section. That is correct behaviour and matches what single-currency stores already experience, but it is a behaviour change on a settings screen rather than a pure display fix, which is why it sits in its own commit with its reasoning.

PR #401, opened 2026-07-29
TWO-25270/fix: make the checkout brand tagline translatable

dgjlindsay:
## Problem

The tagline rendered under the payment-method title at checkout was **whole-sentence brand config** (`checkout_subtitle`), handed to gettext as a **variable msgid**:

```php
wp_kses_post(__($subtitle, 'twoinc-payment-gateway'))
```

gettext extraction is static, so a variable msgid never enters `languages/*.pot` / `*.po`. The string was therefore untranslatable *by construction* — and a brand overlay that authored its tagline in its own language rendered that language on **every** locale, English shops included, with no catalogue entry available to correct it.

The base `two` brand ships an empty tagline and `get_pay_subtitle()` early-returns, so only overlay-branded shops were affected.

## Fix

Invert the ownership of the two halves of the tagline:

- **The sentence is platform copy** — a literal, extractable msgid in `get_pay_subtitle()`:
  `For all companies, %1$sread more%2$s.` with `%1$s`/`%2$s` the opening/closing `<a>` tags.
- **The link target stays brand config** — new brand key `checkout_subtitle_faq_url`. `null`/`''` renders no tagline at all (the Two default, unchanged).
- The former free-form `checkout_subtitle` key is retired; it cannot be made translatable.
- `wp_kses_post` retained (not `esc_html`) — the inline link must survive.
- The URL is escaped *before* the truthiness test, so a scheme `esc_url` rejects (which it returns as `''`) drops the tagline instead of rendering a "read more" that points at the current page; `is_string` keeps a mis-typed array out of `esc_url`, which would fatal the checkout under PHP 8.

Splitting the link into `%1$s`/`%2$s` tag placeholders rather than putting `<a href="…">` in the msgid keeps every brand's URL out of the catalogues and out of translators' hands.

## Also in this PR — English fallbacks in the chip labels (commit 3)

`assets/js/twoinc.js` fell back to the literal `"Payment Terms %s days"` when the translated single-term chip template was missing. An English sentence baked into JS renders as *plausible copy* on a non-English shop and hides the fact that the label never arrived — the same failure class as the tagline above, and it cost an hour of chasing a phantom bug. Both templates now degrade to the shorter localised form and finally to a bare `"%s"`: visibly incomplete rather than convincingly wrong. This mirrors the heading, which already fell back to `''` rather than to English. Two Jest tests cover it; the fallback is defensive only, since `WC_Twoinc_Checkout` is the single localisation site and always supplies both templates.

## Catalogues

| Catalogue | Change |
| -- | -- |
| `twoinc-payment-gateway.pot` | msgid added, alphabetical slot, with a `#. translators:` note |
| `twoinc-payment-gateway-nl_NL.po` / `.mo` | translated; `.mo` recompiled with `msgfmt` |
| `nb_NO`, `sv_SE` | **untouched** — neither carries an equivalent phrase, and both sit 88 msgids behind the `.pot` generally, so this string is left untranslated rather than invented. Runtime behaviour is the English source. |

`msgfmt` reproduces all three committed `.mo` files byte-for-byte from their `.po`, so the `.mo` in this diff is a clean recompile.

## Verification

`make test-unit` all tests passed · `make test-js` 74 passed · `make phpcs` 0 errors · `make phpstan` `[OK] No errors`.

Four adversarial review rounds. Every round after the first found gaps in the *tests* rather than in the fix, and each gap was closed by mutation-testing the guard it was supposed to be:

| Mutation | Now |
| -- | -- |
| reintroduce a legacy free-form branch taking precedence | **fails** `testRetiredFreeFormSubtitleKeyIsInert` |
| blank the nl_NL `msgstr` | **fails** |
| keep the new `.po` but restore the old `.mo` (forgotten `msgfmt`) | **fails** |
| `fuzzy`-mark the entry, recompile (`msgfmt` drops fuzzy) | **fails** |
| revert the JS to the English literals | **fails** the Jest fallback tests (single- and multi-chip) |
| drift the nl_NL msgid from the source literal (msgstr intact) | **fails** |

The catalogue assertion deliberately reads the compiled `.mo`, not the `.po` — WordPress reads only the `.mo`, and the two realistic ways this silently reverts to English are both invisible in the `.po` text.

Rendered `get_pay_subtitle()` output, real code path against the committed `.mo` files, brand URL from the test fixture:

```
en_US: <div class="twoinc-payment-subtitle">For all companies, <a href="https://taglinebrand.example/faq" target="_blank" rel="noopener">read more</a>.</div>
nl_NL: <div class="twoinc-payment-subtitle">Voor alle bedrijven, <a href="https://taglinebrand.example/faq" target="_blank" rel="noopener">lees meer</a>.</div>
nb_NO: <div class="twoinc-payment-subtitle">For all companies, <a href="https://taglinebrand.example/faq" target="_blank" rel="noopener">read more</a>.</div>
```

Not verified on a live shop: this change pairs with an overlay change, and no deploy was performed.

## Known gaps, not fixed here

The escape-before-test ordering above is **not pinned by a test**: the unit harness stubs `esc_url` as identity, so a mutation back to test-then-escape passes. Pinning it needs a stub that rejects a scheme the way real `esc_url` does.

An invalid `checkout_subtitle_faq_url` drops the tagline **silently**. `is_intent_approved_notice_enabled()` sets the opposite precedent in this same file (it logs invalid brand config), so a log line would be consistent — left out to keep this diff to the defect. There is also no i18n gate anywhere in CI (`msgfmt --check`, `.mo`-matches-`.po`, `make-pot` drift); the new unit assertions cover this one string only.

## Breaking change for overlays

An overlay setting `checkout_subtitle` now renders no tagline. Overlays must move to `checkout_subtitle_faq_url`; the paired overlay PR does so.


dgjlindsay:
## Self-review summary — four adversarial rounds

**Round 1** (light pass): no findings — discarded as unreliable, superseded by round 2.

**Round 2** found the fix was sound but the tests were not: the pinned test did not resist reintroduction (a legacy free-form branch taking precedence passed), the user-visible outcome (Dutch on `nl_NL`) was unguarded, and an unusable brand URL rendered an empty `href` or fatalled on a non-string. All fixed in commit 2; escape moved before the truthiness test.

**Round 3** found round 2's own catalogue assertion read the `.po` while WordPress reads only the `.mo` — a forgotten `msgfmt`, or a `fuzzy` marker, still shipped English to Dutch shops with green CI. Fixed by asserting the compiled `.mo`.

**Round 4** found the added placeholder assertion vacuous (nine other entries carry the same placeholders), msgid drift uncaught, and the multi-chip JS fallback unpinned while the commit claimed both branches were covered. All fixed; the `:1647` reference line added in round 3 was reverted as churn — 88 of 218 references in these catalogues are numberless by convention and 75 of the numbered ones are already stale.

**Round 5** found no code defect. Its only findings were a stale test count in this description (introduced by round 4's own new test) and one undocumented design constraint, now recorded in `brands/two.php`: because the sentence is platform copy, every brand setting `checkout_subtitle_faq_url` renders the *same* sentence.

Every guard added across the rounds was verified by mutation, not by inspection — the table in the description lists the mutations and each one fails.

### Deliberately not done

- No i18n gate in CI (`msgfmt --check`, `.mo`-matches-`.po`, `make-pot` drift). The new assertions cover this one string; the gap is repo-wide and predates this branch.
- Invalid brand config drops the tagline silently, where `is_intent_approved_notice_enabled()` logs. Consistent-precedent fix, left out to keep the diff to the defect.
- `assets/js/admin.js` still carries an English `|| "%s days"` fallback — same antipattern, admin-facing, out of scope.

### Not merged

Held for a human call rather than self-merged: round 5's only finding was caused by round 4's fix, which is an oscillation signal rather than a clean round, and the last two commits have not themselves been through a review round. The rendered-output evidence in the description is from the unit harness against the committed `.mo` files — verification on a real shop needs the deploy that follows this merge.

<sub>Review and this comment by Claude.</sub>

dgjlindsay:
## Round 6 — clean round (first-time review of the last three commits)

`7cdd136`, `c032c6e` and `0cc93e3` had never been through a round; rounds 2-5 predate them. Scoped review of just those three found **no defects**.

Verified in this round:

- **Translations ship-worthy.** `bedrifter` (nb_NO) and `företag` (sv_SE) match each catalogue's own register — nb_NO already uses `for bedrifter` collectively and reserves `selskap*` for the counterparty; sv_SE uses `företag` in 15 existing entries. `les mer` / `läs mer` are the conventional read-more link labels. The comma before the link is correct in both languages (a free-standing initial phrase takes one, per Språkrådet and Korrekturavdelingen) and mirrors the source msgid, which is right for fixed platform copy.
- **Placeholders** `%1$s`/`%2$s` present, ordered, single `$`, `#, php-format` on both new entries. `sprintf` round-trips both to a balanced `<a>`. Dropping `%2$s` also fails `msgfmt --check` outright.
- **Encoding** all three `.po` UTF-8, `msgfmt --check` passes, ö/å/ä round-trip, headers and charset undisturbed, no duplicate msgid, and every committed `.mo` is byte-identical to a fresh `msgfmt` of its `.po`.
- **Alphabetical slot** correct in both catalogues (`First name` < `For all companies…` < `Grow your B2B sales…`), same slot as nl_NL.
- **Generalising the catalogue assertion into a loop did not weaken it** — blanking the sv_SE msgstr, drifting the nb_NO msgid, and restoring the old nl_NL `.mo` each fail, with the intended message.
- **Both JS fallback branches pinned** — forcing an English literal into either branch fails exactly the corresponding test. 74/74 Jest.

### Nits accepted, not fixed here

- The locale→msgstr map in the new assertion is hand-maintained: a locale later given a translation gets no coverage, and *removing* one silently narrows the gate. A glob over `languages/*.po` would close both. Folded into the catalogue-lag ticket rather than churned in here, since it belongs with the missing i18n CI gate.
- The new `brands/two.php` comment closes with "brand config cannot carry translatable prose at all", which reads stronger than the file's own precedent — the neighbouring `fee_line_label` key does let a brand supply prose, just untranslatable prose. Factually reconcilable; left alone rather than adding another unreviewed commit for a comment.

Merging base first, then the paired overlay PR — overlay-first would leave the tagline absent rather than wrong-language.

<sub>Review and this comment by Claude.</sub>

PR #402, opened 2026-07-29
TWO-25269/fix: fail closed when the surcharge currency is unquotable

dgjlindsay:
## Problem

On an unresolvable store→checkout FX pair, `build_buyer_fee_share()` returned `null` and the **whole surcharge silently vanished** — the buyer was charged nothing, behind a `warning`-level log and, for the rounds-to-zero case, no log at all. Silently changing what is charged is the worst available outcome.

## Fix: fail CLOSED — withhold the payment method

The check reuses the **existing** gate, `apply_brand_availability_gate()` on `woocommerce_available_payment_gateways`, which already fails closed on FX for the minimum-order gates. No second mechanism.

**Condition** (`WC_Twoinc_Payment_Terms::surcharge_currency_unquotable()`):

> surcharge enabled **AND** store currency !== active checkout currency **AND** at least one offered term carries a `fixed` or `cap` component **AND** `WC_Twoinc_FX::get_rate()` returns `null`

It is deliberately **term-independent**. No term is selected when gateway availability is decided, and it does not need to be — `get_rate()` resolves (or fails to resolve) the pair identically for every term, so "no rate at all" is a property of the currency pair. Gating on "some offered term is unquotable" instead would over-reject a shop with one misconfigured term.

**Placement** is deliberately **outside** the `$basket_is_judgeable` guard and **before** the no-gate-and-no-minimums early return: FX resolvability does not depend on a basket, and the `order-pay` endpoint that guard excludes is exactly a place a surcharge still gets applied. A shop with neither a brand gate nor a minimum still charges surcharges.

## The three rounds-to-zero cases, no longer conflated

| Case | Treatment |
| --- | --- |
| No rate resolvable | **FAIL CLOSED** — gate withholds the method; `error` log |
| `cap` **is** configured and its converted value rounds to `0.00` | **FAIL CLOSED** — no fee block; `error` log. A zero cap is indistinguishable from *no* cap downstream, so relaying it would send an **UNCAPPED** percentage — an overcharge, the opposite of what the merchant configured. |
| `fixed` converted value rounds to `0.00` | **NOT a failure.** Proceeds at `0.00`; `info` log. A legitimately tiny configured fee can be genuinely negligible in a stronger currency and `0.00` is arithmetically correct — rejecting a `0.001` configured fee would be a wrong rejection. |

## An absent cap is still fully legitimate

"No cap defined" continues to charge a non-zero **uncapped** percentage surcharge with the method offered. The fail-closed guard fires only when `$cap !== null` **and** the converted value rounds away. Pinned by `testBuyerFeeShareAbsentCapChargesUncappedPercentage`, which runs the cross-currency conversion path with `$cap === null` and asserts: `surcharge` present, `percentage` present, **no** `cap` key, nothing logged as a failure, method still offered.

## Signals

Every fail-closed path logs at **`error`** level (once per request, matching the gate's own convention — removing a payment method is invisible to the merchant), naming the currency pair and the term. Previously: no-rate logged at `warning`, rounds-to-zero logged nothing.

## Comments and docblocks reversed

- `class/WC_Twoinc_FX.php` — class-header "Fail semantics" block: charge conversions now documented as fail **CLOSED**, pointing at the gate.
- `class/WC_Twoinc_FX.php` — `get_rate()` docblock: "gate **and charge** callers must fail closed on it, display callers fail soft".
- `class/WC_Twoinc_Payment_Terms.php` — the fail-soft rationale block above the conversion; the withheld-surcharge log string; the rounds-to-zero rationale comment (now split into the cap and fixed cases); `fetch_term_fee()`'s "checkout is never blocked" claim and the matching inline comment, both now scoped to **transport errors only**.
- `class/WC_Twoinc.php`'s existing fail-CLOSED gate comment was already correct and is untouched; the method docblock gains the new condition.

## Tests

`tests/unit/run.php` — two existing tests updated to the new semantics and renamed; five added; one shared `assertLogged($level, $needle)` helper, because the log **level** is part of the contract when the failure mode is an invisible removal.

- `testBuyerFeeShareFailsClosedWhenNoRateAvailable` (was `…WithheldWhenNoRateAvailable`) — now also asserts the `error` log
- `testGateWithholdsMethodWhenSurchargeCurrencyUnquotable` — the gate itself
- `testGateKeepsMethodWhenSurchargeIsPercentageOnly` — over-rejection guard; also asserts the FX layer is never consulted
- `testGateFxCheckAppliesOnOrderPayEndpoint` — the `order-pay` placement
- `testBuyerFeeShareCapRoundingToZeroFailsClosed` (was `…WithholdsWholeSurcharge`) — cap-only fixture, asserts the `error` log
- `testBuyerFeeShareFixedRoundingToZeroChargesZero` — `0.00` charged, `info` logged, method still offered
- `testBuyerFeeShareAbsentCapChargesUncappedPercentage` — the item-4 regression pin

`testCartFeeSkippedOnQuoteCurrencyMismatch` was checked and needs no change — it asserts on a pricing **response** echoing the wrong currency, which is unrelated to the FX request-side conversion.

## Gates

`make test-unit`, `make phpcs` (0 errors) and `make phpstan` (no errors) all pass locally.

---

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -2570,9 +2570,11 @@ if (!class_exists('WC_Twoinc')) {
          * Remove the gateway from checkout when the platform minimum
          * (API-resolved, see get_platform_minimum_order()), the brand's
          * billing-country restriction (availability_gate in the brand
-         * config), or the merchant's own minimum is unmet. Mirrors the
-         * brand availability gate semantics: front-end only, minimum is
-         * inclusive (an exactly-minimum basket passes).
+         * config), the merchant's own minimum is unmet, or the configured
+         * surcharge cannot be quoted in the checkout currency at all
+         * (TWO-25269). Mirrors the brand availability gate semantics:
+         * front-end only, minimum is inclusive (an exactly-minimum basket
+         * passes).
          *
          * @param array $available_gateways
          *
@@ -2581,6 +2583,22 @@ if (!class_exists('WC_Twoinc')) {
         public function apply_brand_availability_gate($available_gateways)
         {
             if (is_admin() || !isset($available_gateways[$this->id])) {
+                return $available_gateways;
+            }
+
+            // Surcharge FX resolvability is judged FIRST, and deliberately
+            // OUTSIDE the $basket_is_judgeable guard further down. That
+            // guard exists because the minimums judge a basket; whether the
+            // store→checkout currency pair has a rate does not depend on
+            // the basket at all, and the order-pay endpoint the guard
+            // excludes is precisely a place a surcharge still gets applied.
+            // It also precedes the "no gate and no minimums configured"
+            // early return below: a shop with neither still charges
+            // surcharges. Fail CLOSED, the same shape as the minimums' own
+            // FX handling — better no Two payment method than a buyer
+            // silently charged no surcharge with nobody told (TWO-25269).
+            if (WC_Twoinc_Payment_Terms::surcharge_currency_unquotable($this)) {
+                unset($available_gateways[$this->id]);
                 return $available_gateways;
             }
             $gate = WC_Twoinc_Brand::get('availability_gate');
--- a/class/WC_Twoinc_FX.php
+++ b/class/WC_Twoinc_FX.php
@@ -45,13 +45,17 @@
  * bug). A failed fetch arms a short retry-throttle transient so a
  * flapping API cannot be hammered once per conversion.
  *
- * Fail semantics (per TWO-25104):
+ * Fail semantics (per TWO-25104; charge semantics revised by TWO-25269):
  *  - Gate conversions (minimum-order availability) use last-known-good and
  *    fail CLOSED only when no table has ever been fetched: get_rate()
  *    returns null and the caller must treat null as "cannot be proven".
  *  - Display conversions fail SOFT: callers fall back to the native value.
- *  - Charge conversions (fixed surcharge/cap) fail SOFT to *no* surcharge:
- *    a wrong-currency amount must never be sent to the pricing API.
+ *  - Charge conversions (fixed surcharge/cap) fail CLOSED. A wrong-currency
+ *    amount must never be sent to the pricing API, but dropping the
+ *    surcharge instead charges the buyer nothing and tells nobody, so an
+ *    unconvertible pair withholds the payment method for the whole
+ *    checkout (WC_Twoinc::apply_brand_availability_gate) rather than
+ *    quietly changing what is charged.
  *
  * The response's as_of date (the staleness floor of the rates used) is
  * stored alongside the table for observability and tests; freshness is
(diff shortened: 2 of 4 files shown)


dgjlindsay:
## Adversarial self-review — round 1, no actionable findings

Round 1 raised 1 MAJOR, 4 MINOR and 1 QUESTION. Every one was verified against the fetched ref and **none survived**, so no code changed and there is nothing for a round 2 to oscillate against.

### MAJOR — `get_available_terms()` could return non-iterable and crash the availability filter → **NOT REAL**

`WC_Twoinc_Payment_Terms::get_available_terms($gateway, bool $refresh = false): array` carries a hard `: array` return type, as does `WC_Twoinc::get_merchant_available_terms(bool $refresh = false): array`. A non-array return is impossible — PHP would raise inside the callee, not at the `foreach`. There is no path to a non-iterable here.

### MINOR — no memoisation; the hot filter re-calls `get_rate()` → **NOT REAL**

Two independent bounds, both pre-existing:
- The gate calls `get_available_terms($gateway)` with `$refresh` defaulting to **false**, and the network branch in `get_merchant_available_terms()` is gated on `$refresh && ...`. So the gate performs **no HTTP** — it reads options, which WordPress already caches per request.
- `WC_Twoinc_FX::get_rate()` can trigger at most **one** live re-fetch per request, bounded by the static `self::$fetched_this_request`. Firing the filter 3-5 times cannot produce 3-5 fetches.

Worst case per request is one option read per filter invocation plus at most one FX fetch total. No change warranted.

### MINOR — inconsistent logger guards between the two rounds-to-zero branches → **NOT REAL**

`function_exists('wc_get_logger')` is ANDed into the fixed-rounds-to-zero condition, whereas `log_surcharge_fx_failure()` checks it internally. Different placement, identical outcome: the guarded block only logs, and `$fixed = $converted_fixed;` executes unconditionally afterwards. Logger absence cannot change what is charged on either branch. Deliberately not "aligned" — churning a correct conditional on a payment path for symmetry is not worth the diff.

### MINOR — unknown `$settings['type']` silently yields no monetary components → intended

`surcharge_monetary_components()` returning `['fixed' => null, 'cap' => null]` for an unrecognised type is the safe catch-all: it makes the gate treat the config as currency-agnostic and **not** withhold the method. Failing open on an unparseable type is right here — the alternative is removing Two from checkout because of a config value we failed to recognise.

### MINOR — test log-count assertion is fragile → no action

Correct today. `reset()` clears `__twoinc_test_logs` before every test in the runner loop, so isolation holds.

### QUESTION — FX call ordering / latency hazard → answered above

Both call sites route through the same `$fetched_this_request`-bounded `get_rate()`. Order is irrelevant; there is no second fetch to be had.

### What was verified positively

- **The cap is OPTIONAL and still charges.** The guard is `$cap !== null && $converted_cap <= 0` — the `!== null` limb *is* "a cap is configured", so an absent cap never enters it and falls through to a normal **uncapped** percentage surcharge.
- **`fixed` rounding to 0.00 is not a failure** — it proceeds at 0.00 with an `info` log, method still offered.
- **The gate is term-independent**, keyed only on `get_rate() === null`, so one misconfigured term cannot remove Two for every buyer.
- Every fail-closed path emits an `error`-level log.

CI: 0 failures, 0 pending, `mergeStateStatus: CLEAN`. Leak gate run after committing, three-dot `origin/staging...HEAD` over diff contents, commit body, branch name and PR title/body — zero hits.

### One cross-repo item deliberately NOT fixed here

WooCommerce and PrestaShop currently **diverge** on the cap-rounds-to-zero case: WC drops that term's surcharge and keeps the method offered; PS withholds the payment option. Raised separately for a decision rather than resolved unilaterally — it is a design question, not a defect in either PR, and both are strict improvements on the silent undercharge they replace.

PR #403, opened 2026-07-30
fix(i18n): standardise the PO field label on "PO Number" (TWO-25278)

dgjlindsay:
One phrase for the purchase-order field, **`PO Number`**, on the checkout field label and on the admin toggle alike. Part of a cross-plugin sweep (TWO-25278) that standardises every plugin on the short phrase.

Previously the checkout said `Purchase order number` and the toggle said `Show Purchase order number field`, while the Magento plugin's checkout tile said `PO Number` — so the same field had two names depending on which plugin, and which pane, you were looking at.

Chosen over the longer phrase because it leaves no stale catalogue key behind, carries no cosmetic-breakage risk for shops upgrading, and renders better in the field.

## Surfaces

| file | change |
| -- | -- |
| `class/WC_Twoinc_Checkout.php` | checkout field label |
| `class/WC_Twoinc.php` | settings toggle title |
| `languages/twoinc-payment-gateway.pot` | both msgids |
| `languages/twoinc-payment-gateway-{nb_NO,nl_NL,sv_SE}.po` | both msgids |
| `languages/twoinc-payment-gateway-{nb_NO,nl_NL,sv_SE}.mo` | recompiled |

The toggle keeps the `Show <x> field` pattern shared with the invoice email, project and department toggles, which are untouched.

## i18n

Both msgids are **renamed**, not duplicated, and every locale keeps the translation it already had — all three were already phrased on `PO-nummer`, so nothing needed retranslating:

| locale | `PO Number` | `Show PO Number field` |
| -- | -- | -- |
| `nb_NO` | `PO-nummer` | `Vis feltet for PO-nummer` |
| `nl_NL` | `PO-nummer` | `Veld voor PO-nummer weergeven` |
| `sv_SE` | `PO-nummer` | `Visa fältet för PO-nummer` |

`.mo` files recompiled with `msgfmt` from the edited `.po`. The byte delta on each is exactly the 24 bytes lost from the two shorter msgids, which confirms the same compiler settings as the committed originals. Verified by round-tripping through `msgunfmt`.

Entries are renamed **in place** rather than repositioned. The catalogues are not in any reproducible collation order — neither byte nor case-insensitive sort reproduces the committed sequence — so there is no correct position to move them to, and a re-sort would be pure churn.

## Not changed

Display copy only. `add_field_purchase_order_number`, the `purchase_order_number` field id and order meta, and the `buyer_purchase_order_number` payload key are all untouched.


dgjlindsay:
Adversarial self-review — **round 1 clean, no findings**.

Lenses applied to the diff:

- **missed surfaces** — repo-wide case-insensitive sweep across `php`, `js`, `md`, `po`, `pot` and tests. Remaining occurrences are the lowercase prose ones naming the *business object* rather than the field: the toggle's `description` ("…input their purchase order number to display on the invoice.") and the canonical field-order comments in `WC_Twoinc_Checkout.php`. Deliberately left; they are not the label.
- **catalogue integrity** — both msgids present in the `.pot` and all three `.po`; no duplicates introduced; source references and flags preserved; `msgfmt -c` clean.
- **`.mo` sync** — recompiled from the edited `.po` and round-tripped back through `msgunfmt` to confirm the new msgids resolve. Byte delta on each file is exactly the 24 bytes lost from the two shorter msgids, which is the strongest available evidence that the same compiler settings produced these as produced the committed originals.
- **tests** — nothing under `tests/unit` asserts the label text; the fixtures use the `purchase_order_number` key, which is unchanged.
- **blast radius** — invoice email, project and department toggles byte-identical; `add_field_purchase_order_number`, the order meta key and `buyer_purchase_order_number` untouched.

A second, cross-repo round audited sweep *completeness* rather than this diff — whether any repo or surface was missed entirely, and whether any e2e spec selects the field by its label text.

dgjlindsay:
Adversarial self-review — **round 2 clean**. This round audited sweep *completeness* across all six plugin repos rather than re-reading this diff, since the real risk in a copy sweep is a surface missed entirely, not a mistyped string.

Nothing required a change here. Three things it established that are worth recording:

- **No test anywhere selects the field by its label text.** Every Playwright spec and unit test in the plugin repos locates it by `enable_po_number` / `two_po_number` / `purchase_order_number`, so no suite was silently coupled to the old copy.
- **The three repos claimed to need no change genuinely need none** — `magento-hyva-extension` already read `PO Number`; `magento-abn-plugin` carries no purchase-order label string at all, so its `dev/i18n-audit.sh` cannot trip on the new base msgids; `woocommerce-abn-plugin`'s catalogue entries are dead weight, because it loads only its own textdomain and installs no gettext filter, so a base-plugin string can never resolve through it.
- **Descriptions and help text are deliberately not uniform, and that is not an oversight.** Where the sentence names the *field*, it now says `PO Number` field. Where it names the buyer's own purchase order — "…register their purchase order number", "…input their purchase order number to display on the invoice" — it is left as ordinary prose about the document, because that is what it is describing. Only labels were standardised.

**Out of scope, but Doug should see it:** `PO Number` is now the phrase in all four plugins, but the hosted checkout page, the merchant portal and the invoice PDF still say `Purchase order number`. A buyer who goes through a plugin checkout and then reads the invoice sees both. That gap is not addressable from these repos and is flagged for a follow-up rather than fixed here.

PR #404, opened 2026-07-30
TWO-25269/revert: relay a zero surcharge cap instead of failing closed

dgjlindsay:
## TWO-25269 — revert the zero-cap fail-closed guard

PR #402 added two things under TWO-25269. One was right, one rested on a false premise. This reverts only the false one.

### The false premise

The guard failed **closed** when a configured surcharge cap converted to `0.00` in the checkout currency, on the reasoning that "a zero cap is indistinguishable from _no_ cap downstream, so relaying it would send an UNCAPPED percentage — an overcharge".

That is not the contract. A cap of **zero** instructs the pricing API to clamp the fee to zero. An **absent** cap is what means uncapped. The two are distinct instructions, so relaying a zero cap does not produce an uncapped percentage and there is no overcharge — a cap worth nothing in the checkout currency simply means charge no fee. Verified against the API's own behaviour and test suite under TWO-25269, which holds the references.

Worth noting what the old guard actually did: it withheld the **fee block**, not the payment method. So it reached the *same* zero-fee outcome, just silently and with an error log. Relaying `cap => 0` gets there explicitly, as the merchant's own instruction, and says so at info level.

### Reverted

- The `if ($cap !== null && $converted_cap <= 0)` fail-closed block in `build_buyer_fee_share()` and its log about relaying an "UNCAPPED percentage".
- Every comment and test assertion carrying that premise.

A converted cap of `0.00` now passes through as `cap => 0`. Not dropped, not turned into "no cap", not withheld.

### Also corrected: what the cap actually caps

Review turned up a second error in the same prose, pre-existing and propagated in four places including `assets/js/admin.js`. The cap is **not** a cap on the percentage portion — it bounds the **whole fee line item**, applied after the percentage passthrough and the fixed surcharge are summed. So on `fixed_and_percentage`, a cap of zero zeroes the merchant's **fixed** surcharge too.

That is a strictly larger silent effect than the neighbouring `fixed → 0.00` case, which already logged at info precisely because a fee line quietly reading `0.00` looks like a bug. So a converted cap of `0.00` now logs at info as well, symmetric with it.

### Deliberately kept

- `surcharge_currency_unquotable()` in full, and its wiring into `WC_Twoinc::apply_brand_availability_gate()`. A currency pair with **no rate at all** genuinely cannot be relayed and still fails closed.
- The no-rate branch inside `build_buyer_fee_share()`, `log_surcharge_fx_failure()`, and the `$fx_failure_logged` once-per-request latch. The new cap log deliberately does **not** route through it: that latch is a once-per-request *error* channel, and a zero cap is not a failure.
- `surcharge_monetary_components()`, used by the gate.
- `fixed → 0.00` proceeding at `0.00` with its info log.
- All `round_amt()` call sites. Rounding is untouched — the API rejects sub-cent values, so removing it would produce an HTTP 422, not a more faithful amount.
- Everything from PR #400 (store-vs-active currency).

### One pre-existing gap documented, not fixed

`surcharge_monetary_components()` keeps a cap only when the configured store-currency value is `> 0`, so a cap typed as exactly `0` is normalised to **absent** and relayed as **uncapped** — the opposite of `cap => 0`. Only an FX-converted `0.00` reaches the API as zero. This boundary therefore does not distinguish what the API distinguishes. Pre-existing, deliberately unchanged here, now recorded in the docblock and needing its own ticket.

### Tests

`testBuyerFeeShareCapRoundingToZeroFailsClosed` becomes `testBuyerFeeShareCapRoundingToZeroRelaysZeroCap`: same cross-currency fixture, now asserting `cap === 0.0` is relayed, that the conversion path really ran (`fx_requests > 0`), that exactly one info line is emitted, and that nothing is logged at error or warning. Every no-rate and `fixed → 0.00` assertion kept as-is.

```
$ php tests/unit/run.php     All tests passed.   (147 PASS, 0 FAIL)
$ npm run test:js            3 suites passed, 74 tests passed
$ make phpcs                 exit 0 (0 errors)
$ make phpstan               [OK] No errors
```

### Review

Seven adversarial rounds. The code converged at round 2 and every round since confirmed no over-revert and no under-revert. An explanatory doc added along the way did **not** converge — rounds 5, 6 and 7 each found an error inside prose the previous round's fix had introduced — so it has been dropped, and the findings worth keeping were moved into the comments on the code they describe. Remaining comments cite only the API's documented contract, never its internals.


dgjlindsay:
## Adversarial review summary — 7 rounds, NOT merging, human review requested

Seven review rounds ran against this diff. **I am deliberately leaving this unmerged**, and the reason matters more than the outcome.

### Why it is not merged

The **code** converged early and has been stable since round 2. Every round from 2 to 7 independently confirmed: no over-revert, no under-revert, `surcharge_currency_unquotable()` and its gate wiring byte-identical to `staging`, the no-rate branch intact, `log_surcharge_fx_failure()` and its once-per-request latch intact, `surcharge_monetary_components()` intact, the `fixed → 0.00` info log intact, all `round_amt()` call sites untouched. The only behaviour deletion is the `return null` this PR exists to remove.

The **documentation** did not converge. Rounds 5, 6 and 7 each found an error inside prose that the immediately preceding round's fix had introduced or reworded. That is oscillation, not convergence, so the standing rule is stop and report rather than merge — regardless of CI being green.

### What each round found

| Round | Finding |
|---|---|
| 1 | Nothing. |
| 2 | The cap bounds the **whole fee line item**, not the percentage portion — a pre-existing error this PR was propagating. Missing symmetric info log for `cap → 0.00`. A vacuous gate assertion in the test. |
| 3 | **Verbatim private-repo source, schema text and test identifiers had been copied into this public repo** (see below). Plus a gross-up wording error and a weak log assertion. |
| 4 | The plugin's own `> 0` filter normalises a *typed* cap of `0` to **absent/uncapped** — contradicting this PR's own claim that absence and zero are distinct. Stale "cap bounds the percentage portion" comment still live in `assets/js/admin.js`. `Rounding stays` over-claimed. |
| 5 | Three overstated claims introduced by round 3's and 4's own fixes. |
| 6 | `round_amt()` rounds to `wc_get_price_decimals()`, not a hardcoded 2 — so the rounding guarantee does not hold at 0 or ≥3 decimals. |
| 7 | The doc claimed the reverted guard withheld the **payment method**. It did not — it withheld the fee block, so the surcharge silently vanished while the method stayed on offer. Inflated counterfactual, and self-contradictory against the doc's own backstop paragraph. Corrected here and in the PR body. |

Rounds 2, 3 and 4 were worth having. Rounds 5–7 were the loop.

### Leak disclosure — needs a human decision

Three commits containing **verbatim excerpts from the private pricing-service repo** (the cap clamp expression, the `cap` schema field declaration, its field-description string, a pytest identifier, and internal module paths) were pushed to this **public** repo before round 3 caught them: `7fd80f6`, `3282a6f`, `10f6a2a` — in file content *and* commit-message bodies. The PR was also **created** with a body containing the same quotes.

The branch has since been rewritten so nothing dirty would merge, and the current tip and current PR body are clean — but **a force-push does not remove them.** They stay fetchable by SHA, and the original PR-body revision is retained by GitHub. No further remediation has been attempted: the choice between history rewrite, accept, or changing visibility is not mine to make.

**No partner, merchant, or personal names were involved** — that separate and worse category is clear. (`checkout-api` also appears in `class/WC_Twoinc.php` and two e2e specs in those commits; that is pre-existing on `staging`, not part of this.)

### Verification

```
php tests/unit/run.php   All tests passed.   147 PASS, 0 FAIL
npm run test:js          3 suites passed, 74 tests passed
make phpcs               0 errors (pre-existing line-length warnings only)
make phpstan             [OK] No errors
gh pr checks 404         17/17 pass as of the previous commit; re-running for the final doc-only commit
```

### What a reviewer should actually look at

1. Whether the **revert itself** is right — that a converted cap of `0.00` should reach the API as `cap => 0`. This is the load-bearing decision and it rests on the API clamping on presence rather than truthiness. Source references are on TWO-25269 rather than quoted here.
2. Whether `docs/surcharge-fx.md` should ship at all. It is the entire oscillating surface. Dropping it would leave the revert plus the corrected inline comments, which is a materially smaller and fully-converged change.
3. Three pre-existing gaps this work surfaced but deliberately did **not** fix, each flagged in the doc and worth its own ticket: the typed-`0`-cap asymmetry, sub-cent values on same-currency stores, and `round_amt()` honouring a store decimal setting that is not 2.

<sub>Review conducted by Claude.</sub>

PR #405, opened 2026-07-30
TWO-25283: move "Skip user validation at order confirmation" into Debug Options and explain it

dgjlindsay:
Linear: [TWO-25283](https://linear.app/two-inc/issue/TWO-25283) (parent [TWO-24739](https://linear.app/two-inc/issue/TWO-24739))

## Why

The `skip_confirm_auth` checkbox sat in **Checkout options** with no description, so its title — "Skip user validation at order confirmation" — was the only clue a merchant had about what validation is skipped or what they give up by enabling it.

It is a debug/escape-hatch option, not a checkout feature.

## What it actually does

On the order-confirmation callback the plugin validates two things:

1. the `order_reference` from the request — a per-order `wp_generate_password(64, false, false)` — against the order's stored meta. **Always checked, unaffected by this setting.**
2. a WordPress nonce (`wp_verify_nonce`, key `confirm_<order_id>`). **The only check this setting skips**, and only its *validity*: `is_confirmation_page()` requires the nonce parameter to be present, so with it missing the callback returns early and nothing is processed at all.

The nonce is minted at order-create time. Its window is typically 12–24 hours (`wp_nonce_tick` halves the `nonce_life` filter's `DAY_IN_SECONDS` default, and `wp_verify_nonce` accepts `tick-1`; a site can filter `nonce_life` to anything), and for a signed-in customer it also breaks when their session changes — `wp_create_nonce` binds to the user ID and `wp_get_session_token()`. On a guest checkout both of those are empty, so there, expiry is the only trigger. Either way the buyer is shown "The security code is not valid." on a legitimately authorised order, and this setting is the way out.

The cost is the nonce's protection against a cross-site request. The 64-character order reference still has to match, so the callback is not left open.

## Changes

- `skip_confirm_auth` moved into the `section_debug` group, next to `enable_api_logging`.
- Real `label` (was the `' '` placeholder) plus a two-sentence `description` covering all three points: what is skipped, why a merchant might need it, what it costs.
- Default unchanged (`no`), stored option key unchanged (`skip_confirm_auth`).
- New msgids translated in `nl_NL`, `nb_NO`, `sv_SE`; `.mo` recompiled, `.pot` updated.

## Verification

**Rendered grouping.** WooCommerce renders `form_fields` in array order, with `type => title` entries opening a section, so position is the grouping. Walking the built field array gives:

```
## section_payment_terms — Payment terms and offset pricing
   ... surcharge_rounding_basis, surcharge_rounding_step

## section_debug — Debug Options
   enable_api_logging
   skip_confirm_auth
```

`testSkipConfirmAuthRendersUnderDebugOptions` asserts that walk, and pins one field known to live elsewhere (`display_tooltips` → `section_checkout_options`) so the assertion cannot pass vacuously. Reverting the source change makes it fail.

**Stored value survives the move.** `WC_Settings_API::init_settings()` reads one serialised option row keyed by field name, and `process_admin_options()` skips `title` fields entirely — so section membership is never stored and cannot be orphaned by a move. `testSkipConfirmAuthStoredValueSurvivesSectionMove` drives that through `get_option()`, the same call the confirmation callback makes: it seeds the row as an upgraded install has it (`skip_confirm_auth => 'yes'`), reads it back, checks a never-configured shop still resolves the `no` default, and runs the dead-key sweeper `drop_removed_settings()` over the row. Its docblock states plainly that it *cannot* prove the section is what preserved the value — nothing stores the section — and that what it really catches is a rename or an accidental drop of the field.

That test needed the harness's `WC_Payment_Gateway::get_option` shim to stop ignoring `$settings` and mirror core (stored row first, `get_field_default`'s `empty() => ''` semantics, `$empty_value` for `''`, all memoised). That is a shared-fixture change: about a dozen unrelated tests now resolve real field defaults where the old shim gave them `''`, so their rendered output moved even though the suite stays green. The one deliberate divergence from core — no lazy `init_settings()` — is commented in place.

**Translations.** `testSkipConfirmAuthCopyIsTranslatedInEveryLocale` reads both msgids off the live `form_fields` entry rather than retyping them, asserts they are present in the `.pot`, and asserts each compiled **`.mo`** (not the `.po`) carries both the msgid and a fragment of its msgstr. Asserting the `.mo` is the convention TWO-25270 set: WordPress reads only the `.mo`, and both silent-revert-to-English failure modes — a forgotten `msgfmt`, a fuzzy marker — are invisible in `.po` text. Both regressions are demonstrated to fail the run: editing the source string without regenerating, and restoring a stale `.mo`.

What that does **not** catch, and the docblock says so: the source and all three msgstrs being changed together in the wrong direction. Review found this by deleting a clause from the source and from every msgstr and regenerating — the suite stayed green. Nothing pins the copy's content; that stays a review job.

`php tests/unit/run.php` — all tests pass. CI runs the 7.4–8.4 matrix; 7.4, 8.0, 8.4 and 8.5 also verified locally.

## Review history

Round 1 found nothing. Round 2 found the copy wrong on three claims and one test that passed with the source reverted. Round 3 found that the replacement `.mo` test still compared a hardcoded literal (so source drift stayed invisible), that the shim was less faithful to core than its comment claimed, and that round 2 had over-corrected the copy by dropping the session cause outright. Both fix commits carry the detail. Round 4 found no new medium or high findings and confirmed the two earlier fixes hold, leaving five low/nit items — the blind spot above, three comment inaccuracies, and a defensive array read — all addressed in `TWO-25283/chore`. Round 5 pending.

## Notes, not changed here

- The partner overlay's field-suppression filters are all key-based `unset()`s and do not include `skip_confirm_auth`, so the move does not interact with them.
- `skip_confirm_auth` is now the last entry in `form_fields`, so anything an overlay appends via `apply_filters('wc_two_form_fields', …)` renders under "Debug Options". Pre-existing property of the array's tail (`enable_api_logging` was previously last) and no in-repo overlay appends; noted rather than addressed.
- `WC_Twoinc_Helper`'s `$skip_nonce` parameter has no caller that ever passes `true`, and if one did the confirmation URL would omit the nonce parameter entirely and `is_confirmation_page()` would never fire. Predates this branch; left alone.
- The description is 549 characters, about twice the file's next-longest `description`. Deliberate — the three things a merchant needs to know here do not compress further — but flagged as a style call rather than settled.
- Magento and PrestaShop have no equivalent merchant-facing setting: PrestaShop's confirmation controller validates an attempt token plus a cart snapshot hash with no opt-out, and Magento has no such toggle. Cross-plugin consistency noted, not changed under this ticket.


dgjlindsay:
## Adversarial self-review — five rounds, closing summary

Posting this before merge under the self-review lane (`reviewDecision` is empty on this repo, so no separate reviewer is enforced). A human is welcome to jump in.

**Round 1** — no findings.

**Round 2** — four medium, five low. The copy was wrong on three separate claims: it said "without a WordPress nonce" when `is_confirmation_page()` requires the nonce *parameter* and only its validity is skipped; it blamed a different browser or a changed session when `wp_create_nonce` binds to a user ID and session token that are both empty on a guest checkout; and "expires after about a day" understated a 12–24 hour window. It also overstated the security cost by implying the order reference was the *only* remaining guard rather than a 64-character per-order secret that is still mandatory. Separately, `testSkipConfirmAuthStoredValueSurvivesSectionMove` passed with the source change reverted — it re-read a value it had seeded three lines earlier, because the harness's `get_option` shim ignored `$settings` entirely.

**Round 3** — the replacement `.mo` test still compared a *hardcoded copy* of the literal, so the likeliest regression (source copy edited, catalogues not regenerated) stayed invisible; deleting the description outright left the suite green. The shim was also less faithful to core than its own comment claimed, and round 2 had over-corrected the copy: for a signed-in customer both nonce bindings *are* live, and on a B2B gateway that is a common case.

**Round 4** — no medium or high. Five low/nit: a `setAccessible` version gate that was one of eleven sites and implied a deprecation-cleanliness it did not deliver; a shim comment claiming one divergence from core where there are three; an unguarded array read; and — the substantive one — the blind spot the round-3 fix bought. Reading msgids off `form_fields` means the source and all three msgstrs moving *together* in the wrong direction is now invisible. Demonstrated by deleting a clause from the source and every catalogue and regenerating: green. That is documented in the test's docblock rather than papered over. Nothing pins the copy's content; that stays a review job.

**Round 5** — no findings on the diff. Two plumbing blockers, both now resolved: the branch had gone `CONFLICTING` against `staging` on the three compiled `.mo` files, and the previous head had never actually been through CI.

The `.mo` conflict is worth recording, because resolving it the obvious way would have shipped a silent regression. `staging` had just *removed* a msgid; this branch *adds* two. The `.po`/`.pot` sides auto-merge, the binaries cannot, and picking either side would have been wrong — keeping this branch's side resurrects the removed string in all three compiled catalogues, with no test anywhere to catch it. Resolved by taking the merged `.po` and regenerating all three `.mo` with `msgfmt`, then verifying both directions: the removed msgid is absent from every `.po`, `.pot` and `.mo`, the two new ones are present in all of them, and each `.mo` is byte-identical to a fresh compile of its `.po`.

### State at merge

- All 17 checks pass on the rebased head; `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.
- Suite additionally verified locally on PHP 7.4, 8.0, 8.4 and 8.5; CI covers 7.4–8.4.
- `phpcs` clean (+2 non-gating line-length warnings from the new description), `phpstan` clean.
- Two style calls left deliberately open rather than silently settled: the description is roughly twice the length of the file's next-longest, and the en-dash/em-dash split across locales follows each catalogue's existing precedent rather than being unified.

Merging with `--merge` (this repo disallows squash and rebase). No `--admin`.

<sub>Review and this summary by Claude.</sub>

PR #406, opened 2026-07-30
TWO-25279/refactor(surcharge): enforce a merchant tax rule for the surcharge tax treatment, no grandfathering

dgjlindsay:
## What

**Never taxed** (stored value `always_zero`) is a plugin-supplied never-taxed treatment, not a tax rule the merchant set up. Selecting it silently means "the fee is untaxed for every destination" — a tax decision made by picking an option we handed them.

It is now suppressed **unconditionally**. A merchant who wants an untaxed fee creates a WooCommerce tax class with a 0% rate and selects it under **Specific tax class**, so the treatment is visible and auditable under WooCommerce → Settings → Tax. Matching changes land in the PrestaShop and Magento plugins.

## Enforce, do not grandfather

There is **no** carve-out for a shop that already stores `always_zero`, and no silent rewrite of its settings. Such a shop:

1. does **not** get the option back in its dropdown;
2. is **refused** at save — on the treatment field, with a message naming the consequence, and independently on the surcharge-type field, which will not leave surcharges enabled against a treatment the shop is not offered;
3. sees a **loud error in the field's visible description** ("this store is set to leave the surcharge UNTAXED for every destination") until it picks a real treatment.

Point 3 is what makes point 1 safe. A `<select>` cannot render a value absent from its options, so without the error the field would simply look *unset* while every order was in fact being charged an untaxed fee — the silent outcome, which is worse than either the block or the warning alone. The field's help text moves into the tooltip so the error can occupy the visible slot, which is the shape the neighbouring **Surcharge Tax Class** field already uses for its stale-selection warning.

All four sites — the option list, both validators and the notice — delegate to **one** predicate (`is_never_taxed_surcharge_treatment`, over a named constant), so a shop cannot end up warned-but-savable or blocked-but-unexplained.

## Two details worth a reviewer's eye

**The predicate matches by exact string and deliberately does not trim.** It must be true exactly when the runtime resolver would in fact leave the fee untaxed, and that resolver compares untrimmed against the same canonical value. A settings row carrying stray whitespace (only reachable from a direct DB or API write) therefore resolves to the **Standard** treatment at checkout — so it is refused as *unoffered* rather than reported as untaxed, because claiming it is untaxed would be false. Both behaviours are pinned by tests.

**The notice reads the raw settings row, not `get_option()`.** It is called from `init_form_fields()`, which the constructor runs *before* `init_settings()`, so `$this->settings` is still empty and `get_option()` would silently report no stored treatment — dropping the one warning this exists to raise. There is a dedicated test whose gateway has an empty injected option layer and a populated raw row, so a revert to `get_option()` is caught.

## No migration

Deliberately none. Nothing is rewritten: a shop still holding the mode keeps charging an untaxed fee — visibly — until it chooses. WooCommerce has no version-gated migration runner and none is added.

Also drops the grandfathering-era `trim()` in the runtime resolver, which existed only to keep the admin display and the cart in step while a whitespace-padded row was still renderable in the dropdown. It no longer is, so that file returns to its `staging` content.

## Verification

`php tests/unit/run.php` — **all tests passed** locally. `phpcs` and `phpstan` are not runnable in this checkout (no `vendor/`); both are gated by CI on this PR.

Every enforcement site was mutation-tested — production code deliberately broken, test confirmed red:

| Mutation | Killed by |
|---|---|
| predicate always returns `false` | `a shop that never stored the never-taxed mode cannot newly post it` |
| option list drops the shared-predicate filter | `a shop that never chose is not offered the never-taxed mode` |
| treatment validator drops the never-taxed refusal | `the refusal names the consequence, not just "one of the offered modes"` |
| notice always returns `''` | `a shop storing the mode must be told` |
| notice wired to the tooltip instead of the visible slot | `the notice must be the visible field description, not the tooltip` |
| notice reads `get_option()` instead of the raw settings row | `the notice must read the raw settings row, not $this->settings` |

## Concurrency

`class/WC_Twoinc.php` is a contended file. This revision has **not** been re-trial-merged against the other in-flight branches — the earlier revision's clean trial-merge result no longer applies to this diff. The hunks here are the new methods and constant near the top of the class, one line in the `surcharge_tax_treatment` field definition, and the two validators.

## Translations

`languages/twoinc-payment-gateway.pot` carries **no** entry for the surcharge tax treatment title, its help text, or any of its option labels — the whole tax-treatment block was never added to the catalogues. So no translation is orphaned, and no `.po`/`.mo` is regenerated here. The two new error strings are untranslated for that same pre-existing reason. That Norwegian/Dutch/Swedish gap predates this change and is worth closing separately.


dgjlindsay:
## Adversarial self-review — round 1: not clean, 4 major (no blocker). All addressed in `adb9132`.

### `get_option()` was the wrong read inside `init_form_fields()`

The option builder called `$this->get_option()`, but the constructor runs `init_form_fields()` **before** `init_settings()`, and `WC_Settings_API::get_option()` lazily re-enters `get_form_fields()` when `$this->settings` is empty. On a shop with no settings row that primed `$this->settings` from an empty field list; it survived only because the constructor overwrote it one line later.

That is exactly the ordering the sibling `get_checkout_env_options()` already avoids, with a comment saying why. The builder now reads the raw settings row the same way. This matters beyond tidiness: a half-primed `$this->settings` reports **no stored treatment** and drops the option — precisely the lockout the carve-out exists to prevent.

Round 1 did confirm the lockout rationale itself against real WooCommerce source: `process_admin_options()` catches a per-field `Exception` and `update_option`s the `$this->settings` seeded by `init_settings()`, so a throwing field keeps its old value and the shop sits on a permanent admin error. (It also found that the *documented* infinite-recursion claim at `WC_Twoinc.php:231-235` is stale — current `get_form_fields()` does not call `init_form_fields()`. The ordering problem is real regardless.)

### The rule was UI-only

`validate_surcharge_tax_treatment_field` still accepted `always_zero` against a hardcoded three-mode list, so a crafted POST persisted it on a shop that never stored it. It now validates against `get_surcharge_tax_treatment_options()`, mirroring `validate_surcharge_tax_class_field` directly below it. Because the list is built from the pre-save stored value, that single predicate expresses the whole policy: a shop already storing the mode can resubmit it, nobody else can newly pick it.

### The option is keyed on the raw stored value

Keying on the literal `'always_zero'` while matching on `trim()` meant a row of `' always_zero '` was matched yet still unrenderable — a `<select>` can only render an option whose key is byte-identical to its value, so that shop hit the lockout anyway. Now keyed on the raw value, as `get_checkout_env_options()` does.

### A test that could not fail

The new spec drove the harness's injected option array, which the production code no longer reads, so it proved wiring rather than behaviour. It now drives `$GLOBALS['__twoinc_test_options'][$gateway->get_option_key()]` — the raw row. An existing assertion that `always_zero` validates was inverted, since it is now correctly refused for a shop that never stored it.

### Mutation testing

Every new assertion was verified by breaking the production code and confirming the test went red — 9 mutations, all killed:

| mutation | killed by |
|---|---|
| always offer the never-taxed mode | `a shop that never chose is not offered the never-taxed mode` |
| never offer it | option-list equality in the already-stored case |
| field hardcodes the option list | `the settings field must render the scope-aware list, not a hardcoded one` |
| help text re-advertises `Never taxed` | `the help text must not offer the never-taxed mode as a choice` |
| treatment validator drops `always_zero` | `resubmitting the stored mode must not be rejected` |
| surcharge-type gate drops `always_zero` | `the surcharge-type gate must accept the stored never-taxed mode` |
| read via `get_option()` instead of the raw row | `the option must be keyed on the raw stored value, whitespace included` |
| key on the literal instead of the raw value | same |
| validator back to the hardcoded three modes | `a shop that never stored the never-taxed mode cannot newly post it` |

### Confirmed clean

`assets/js/admin.js` switches on `treatment === "custom_class"` — value-based, no `always_zero` reference, no ordering or index dependence. `WC_Twoinc_Payment_Terms` still handles `always_zero` at runtime, which is correct: legacy shops keep working. Multisite and brand overlays are fine — `$this->id` is set before `init_form_fields()`, so `get_option_key()` resolves the right per-brand, per-site row. PHP 7.4 fine. The WC REST `update_settings` path already checks `$field['options']` membership, consistent with the new policy. No `.pot` / `.po` obligation: none of these admin strings are in the catalogues, and that gap predates this change.

### Left as-is, flagged

`tests/unit/bootstrap.php`'s `WC_Payment_Gateway::get_option()` stub returns `$empty_value ?? ''` unconditionally and never mirrors WC's lazy `init_settings()`, so no test can exercise that ordering. The raw-read fix removes this code's dependency on that ordering entirely, which is the better answer than teaching the stub; changing the stub would also ripple across the whole suite. Worth a separate ticket if other code grows the same dependency.

### Concurrency note

`class/WC_Twoinc.php` is contended. This branch was trial-merged against `origin/doug/TWO-25283-skip-confirm-auth-debug` (PR #405) before the first push — `class/WC_Twoinc.php` and `tests/unit/run.php` both auto-merge cleanly — and `origin/staging` has since been merged in, picking up #404 and #407.

*Review by Claude.*

dgjlindsay:
## Adversarial self-review — round 2: **not clean, and oscillating.** Holding for human review.

Round 2 found a **BLOCKER that round 1's own fixes created**. Per the merge policy that is an oscillating loop, not progress, so this PR is **not being auto-merged**.

### BLOCKER (caused by round-1 fixes 2 + 3) — a padded row could not be saved at all

Round 1 did two things independently:

- **fix 3** keyed the re-offered option on the **raw** stored value, so a `' always_zero '` row could render at all;
- **fix 2** made `validate_surcharge_tax_treatment_field` check option membership **after** `trim()`ing the posted value.

Each is correct alone. Together, for a stored `' always_zero '`, **both** the raw value the select posts and its trimmed form threw *must be one of the offered modes* — the shop could not save anything. That is precisely the permanent-admin-error lockout the carve-out was added to prevent.

Fixed by registering **both** keys. The canonical `'always_zero'` is what every validator and the runtime resolver compare against; the raw form sits alongside it purely so the select can render the row. Because the validator returns the trimmed value, the odd row **normalises itself on the next save**.

### MAJOR — the two validators had diverged, and it was exploitable

`validate_surcharge_type_field` still gated on the hardcoded `['standard','custom_class','always_zero']` while the treatment validator used the offered list. `surcharge_type` precedes `surcharge_tax_treatment` in `init_form_fields` and WooCommerce validates in that order, so a POST of `surcharge_type=percentage` + `surcharge_tax_treatment=always_zero` on a shop that never stored it: the type gate **passed**, the treatment field threw, and WooCommerce kept the old (blank) treatment. Net saved state: **surcharges enabled with no treatment** — the exact invariant that gate exists to enforce. Both validators now share one source.

### MAJOR — a padded row rendered a lie

`WC_Twoinc_Payment_Terms::resolve_surcharge_tax_treatment` did not trim, so `' always_zero '` resolved to `standard` at checkout while the new carve-out rendered that row as a selected **Never taxed**. The admin was claiming a treatment the cart was not applying. Now trimmed.

### MINOR — the round-2 test asserted the render, never the round trip

The padded case stopped at `array_key_exists(' always_zero ', …)`, which is exactly why the blocker above passed CI while every other branch of that test walked the full save path. It now posts the value back through both validators, asserts the canonical key is registered too, and asserts the runtime resolution.

### Flagged, not fixed (pre-existing)

`WC_REST_Payment_Gateways_V2_Controller::update_item` validates only via `validate_setting_<type>_field` against `$field['options']` — it never calls `validate_surcharge_type_field`. So REST can persist `surcharge_type=percentage` with an empty treatment. The **suppression itself is** enforced over REST, because that path reads the new scope-aware `options`; the cross-field type gate is not. Pre-existing, wider than this ticket, worth its own issue.

### Verification after the round-2 fixes

`php tests/unit/run.php` green, `phpcs` (4.0.1) 0 errors, `phpstan` (2.2.5, pinned stubs) no errors. Round-2 changes mutation-tested: dropping the canonical key, dropping the raw key, reverting the type gate to the hardcoded list, and removing the resolver's trim were each broken in turn and each went red.

### Round 2 confirmed clean on

Pre-save state is correct for the option builder — WC's `process_admin_options` validates every field before `update_option`, and REST likewise calls `init_form_fields()` before touching `$settings`. `get_option_key()` is stable when `init_form_fields()` runs (`$this->id` is set first). `get_option` (not `get_site_option`) is the right multisite scope for gateway settings. No key-collision or escaping path: `trim($stored) === 'always_zero'` admits only whitespace variants, which cannot collide with `''`/`standard`/`custom_class`, and WC `esc_attr`s option keys. `assets/js/admin.js` does not assume the three-mode list. PHP 7.4 compatible. `drop_removed_settings` / `drop_renamed_option_rows` do not touch the treatment row.

*Review by Claude.*

PR #407, opened 2026-07-30
TWO-25287/fix: remove the invoice-email hint on the checkout

dgjlindsay:
## What

The optional invoice-email field on the checkout carried a brand-interpolated placeholder, `"Only for invoices being sent by %s"`. It was never asked for, and none of its sibling optional fields (PO number, project, department) has ever had one.

- Drop the `placeholder` key entirely from `$fields['billing']['invoice_email']` in `class/WC_Twoinc_Checkout.php`, so the field renders like its siblings.
- Remove the now-dead catalogue entries from `.po` (nb_NO / nl_NL / sv_SE) and `.pot`; `.mo` files recompiled with `msgfmt`, not hand-edited (`msgfmt` reproduces the committed `.mo` byte-for-byte from the unmodified `.po`, so the only `.mo` delta is the removed entry).

Cross-platform counterpart of TWO-25281, which removed the same copy from the PrestaShop checkout.

## Sole consumer

`class/WC_Twoinc_Checkout.php` was the only place the string was used. The pay-for-order view (`views/woocommerce_order_pay.php`) re-renders company/checkout inputs with its own inline JS but does not render the invoice-email field, so nothing else regresses.

## Test

New `testInvoiceEmailFieldHasNoPlaceholder` asserts the field is still added when `add_field_invoice_email` is on, and that it defines no `placeholder` key.

**Verified by mutation:** reinstating `'placeholder' => 'MUTATION'` on the field makes the test fail with `invoice_email must not define a placeholder`; removing it again returns the suite to green (`php tests/unit/run.php` → `All tests passed.`).

## No version bump

Display-string only — persists nothing, no upgrade script.


dgjlindsay:
## Self-review — clean round

**Scope:** one deleted `placeholder` key, four catalogue files, one new test.

Checked and found nothing:

- **No other consumer of the string.** Repo-wide grep for `Only for invoices` returns only the catalogue entries this PR deletes. The pay-for-order view (`views/woocommerce_order_pay.php`) re-renders phone / country / company-name / company-ID inputs with its own inline JS, but never the invoice-email field, so it is unaffected.
- **Catalogue integrity.** Each entry removed as a whole block (reference comment, `php-format` flag, `msgid`, `msgstr`) with no orphan comment lines. `msgfmt -c` clean on all three `.po` files.
- **`.mo` correctness.** `msgfmt` reproduces each committed `.mo` byte-for-byte from its *unmodified* `.po`, so the binary delta here is exactly the removed entry and nothing else — no hand-editing, no toolchain drift smuggled in.
- **Test is not vacuous.** It pins both halves: the field is still added when `add_field_invoice_email` is `yes`, and it defines no `placeholder` key. Mutation-verified — reinstating a placeholder value fails the suite with `invoice_email must not define a placeholder`.
- **No unused symbol left behind.** `WC_Twoinc_Brand` is still referenced elsewhere in the file.

CI green across phpcs, phpstan, the six-PHP-version unit matrix, Jest, boot smoke, upgrade smoke and e2e.

---
*Review by Claude.*

PR #408, opened 2026-07-30
fix(TWO-25288): paint the company-search spinner from an animated GIF

dgjlindsay:
## What

TWO-25288, element 1 only. Gives the company-search field an in-field loading spinner, painted
from an animated GIF — the same loading GIF the PrestaShop plugin already ships, copied into this
repo at `assets/images/loader.gif`.

Before, the indicator borrowed the shared three-dot pulse styling through a nested three-element
structure. Now it is a single childless `<span>` with `aria-hidden="true"`, and the stylesheet
paints the GIF onto that one node. No inner markup for the stylesheet to stay in step with.

## Why an animated image rather than CSS

**An earlier revision of this branch drew the spinner in CSS and it did not turn.** A stepped
timing function whose increment matched the spoke pattern's period mapped the gradient onto
itself at every step, so it rendered completely motionless — and shipped that way for a commit
with the whole suite green. jsdom does not evaluate animations, so neither the figure's geometry
nor its timing was verifiable in CI at all. That is not a gap you can close with a better
assertion; it is a property of the medium.

Using an animated image makes the motion a property of the asset instead of a
gradient-and-timing interaction only a real browser can judge.

All the CSS apparatus that supported the drawn figure is gone: the repeating conic gradient, the
mask pair, the border-radius, the em-relative sizing, the keyframes block and the colour
variable. The `prefers-reduced-motion` rule goes too — CSS cannot pause, slow or step an
animated GIF, so that rule was a literal no-op that would have misled the next reader into
thinking the preference was honoured.

## Implementation notes

**`background-image`, not an ``.** A `background-image` URL resolves relative to the
stylesheet, so no asset URL has to be plumbed through the platform's asset helper and into the
markup, and the node stays childless — which is exactly what the background needs.

**Fixed 16px, not `1em`.** The asset is natively 16x16; an em-relative box would upscale and blur
it.

**The selector stays a single flat class,** which is cheap and preserves the styling hook. It no
longer claims a brand edition can recolour the spinner, because an overlay cannot recolour a GIF.

## Tests

The existing markup assertions (childless, `aria-hidden`, exactly one node, removed rather than
hidden when the search ends, lands in the widget's own search box) all still hold and are kept.

**Added one test that observes the paint,** because nothing pinned that the asset resolves at
all — and markup assertions stay green whether the node paints anything, which is precisely how
the motionless spinner shipped. It injects the real stylesheet into the document as a `<style>`
element (jsdom does not fetch link hrefs), drives a real search, and reads back the computed
`background-image`. Paired with an on-disk existence check, because a correct URL aimed at a
missing file satisfies the computed style on its own.

The stylesheet-injection helper lives in the harness rather than inline in the test, and is
documented there and in the test README.

**Mutation-verified.** Each of these turns the new test red on its own, with the rest of the
suite green:

| Mutation | Result |
| --- | --- |
| point the CSS `url()` at a nonexistent filename | RED |
| delete the `background-image` declaration | RED |
| delete `assets/images/loader.gif` | RED |

Full suite: `79 passed, 3 suites`. `pre-commit run --all-files`: prettier passed.

## Honest bound on what the suite proves

jsdom's cascade is partial. It resolves `background-image`, `background-repeat` and
`background-size`, so the rule can be proven to point at the asset. It does **not** resolve the
multi-value `background-position` shorthand — that reads back empty however the rule is written,
so it is not asserted. And nothing here proves the result is *visible* or that the image
*animates*: box geometry, stacking and animation are all beyond jsdom. Treat a change to the
spinner's appearance as needing a real browser. Recorded as such in `tests/js/README.md`, whose
previous account of the motionless-spinner near-miss described code that no longer exists and has
been rewritten.

## Scope

Element 1 only. No version-file change — this repo derives the version in CI from the PR's
conventional-commit types. No new buyer-visible strings, so no translation-catalogue change.

Rebased onto `staging` (it had moved on); clean, no conflicts.

## Review status

**Not ready to merge — round 3 of adversarial review has not run.**

---


dgjlindsay:
## Adversarial review — three rounds, third one clean

Three independent lenses with non-overlapping mandates: correctness/runtime edge cases, platform idiom, and tests/maintainability. Every finding was re-verified against the fetched ref before being acted on.

**Round 1 — one MAJOR finding here, fixed in `d1eca5c`:** the "painted with the animated loading image" test resolved the asset from a **hardcoded path** rather than from the `url()` the stylesheet actually declares — so repointing the URL at a different directory still passed, while the test's own comment and `tests/js/README.md` both claimed it caught exactly that. Proven by mutation: `url("../imagez/loader.gif")` left the suite fully green at 79/79, whereas the identical mutation in the sibling Magento repositories went red. The test now parses the declared `url()` out of the computed `background-image`, resolves it against the stylesheet's own directory, fails loudly if the parse misses, and asserts on the resolved file's bytes.

**Round 2 — one MAJOR finding, caused by round 1's own fix, fixed in `3902322`:** the new frame count scanned the whole file for raw `0x2C` bytes. That value occurs in colour tables and compressed pixel data, so a genuinely single-frame GIF passed — the exact "spinner that never spins" regression the assertion was added to catch. A valid 242-byte single-frame GIF89a carved from this very asset contains two `0x2C` bytes and sailed through. Replaced with a real GIF block-structure walk in `tests/js/wc-harness.js`: header, global colour table, extension sub-blocks, image descriptors with local colour tables, trailer.

Round 3 exercised that parser rather than reading it: 12 frames on the shipped asset, 1 on two independently-built single-frame GIFs, 3 on a synthetic three-frame GIF stuffed with local colour tables, interlacing, a `NETSCAPE2.0` extension, a comment extension, `0x2c`-filled colour tables and trailing garbage after the trailer, and correct termination on 12 adversarial inputs including all-`0xFF`, all-`0x2C` and truncated buffers. 20000 random buffers produced zero divergence from the PrestaShop copy.

**Mutations verified red, each restored afterwards:** repoint the `url()`; delete the `background-image` declaration; delete `assets/images/loader.gif`; substitute a valid single-frame GIF. The last two are the ones that used to pass.

**Round 3 — CLEAN.** One finding, and it is explicitly independent of every fix made in this loop: both `countGifFrames` walkers return a partial count for a *truncated* GIF instead of signalling malformation, so a corrupted asset with two or more recoverable image descriptors would still satisfy `> 1`. The same weakness existed in the raw byte scan this loop replaced, so it is pre-existing rather than introduced. Filed as minor and deliberately not fixed — a `-1`/throw when the walk ends without reaching the `0x3B` trailer is the fix if it is ever wanted.

**Not oscillating.** Round 2's findings were not all attributable to round 1's fixes (PrestaShop carried an independent twin of the unsound frame count), and round 3's single finding was caused by neither round. The loop converged.

**Verified at source, not taken on trust:**

- The asset is byte-identical in all four plugin repositories (`md5 03ce3dcc84af110e9da8699a841e5200`, 1553 bytes), is a genuine `GIF89a` 16x16, and carries **12 frames at 70ms with a `NETSCAPE2.0` loop count of 0** — it loops forever rather than running once and stopping.
- The `url()` in each stylesheet resolves to a file that exists on the branch, relative to that stylesheet's own published location.
- No packaging step drops the asset — checked `.gitattributes` / `.distignore` / the release script in each repository.
- Nothing removed by this change was still live: the payment-term chips' dot rules and keyframes all still have consumers, and `prefers-reduced-motion` has zero occurrences left anywhere, so deleting it orphaned nothing.

**Left open, and worth knowing before this is looked at on a shop:** the rendered result has not been seen in a browser. Specifically, whether `url("../images/loader.gif")` resolves after `setup:static-content:deploy` on the two Magento surfaces is argued from the layout registration and the static-content directory layout, not from an observed page.

---
*Review by Claude.*

PR #409, opened 2026-07-30
TWO-25289/fix(surcharge): refuse a cap of 0, relay a stored 0 verbatim, round relayed amounts to 2dp

dgjlindsay:
## TWO-25289 — a surcharge cap of 0

Three related things at the surcharge boundary, plus one CSS nit.

### 1. Refuse a cap of `0` at entry

A cap of exactly `0` is never what a merchant means by it.

- The cap bounds the **whole fee line** — the percentage part and the fixed fee together, not the percentage alone. So a cap of `0` silently wipes a configured fixed fee too, with nothing in the grid saying so.
- The intent it gets mistaken for — "charge nothing on this term" — is expressible directly, by entering 0 in the percentage and fixed cells.

`validate_two_surcharge_grid_field()` now rejects it and says exactly that. An **empty** cap stays valid and still means "no cap": the blank-cell `continue` returns before this check, so absence and zero stay distinguishable. `0`, `0.00`, `0,00` and `00` are all the same value typed differently and all rejected.

No client-side mirror: this grid has no client-side validation idiom to add one to, and inventing one was out of scope. The server is the only gate, which is also the only one that matters.

### 2. Remove the `> 0` filter on the relayed cap

This was the actual overcharge, and it is the reason the ticket exists. `surcharge_monetary_components()` normalised a cap of `0` to **absent** — and absence is what means "uncapped" — so a merchant who configured `0` got an **uncapped** percentage.

Grid validation alone does not close it. A `0` can arrive by any route the grid does not police: a row stored before that validation existed, `wp option update`, programmatic config, an import. Those would still have been relayed uncapped. So a `0` is now relayed as `cap => 0`, which bounds the fee at zero — the merchant's own instruction, explicitly.

The `fixed` component keeps its `> 0` filter deliberately: an absent fixed fee and one of `0.00` are the same instruction, so nothing is lost by conflating them.

### 3. Round monetary values to 2dp before the request

`cap` and `surcharge` went out **unrounded** on the same-currency path, and through `round_amt()` — i.e. `wc_get_price_decimals()`, so 3 or 4 on a store configured that way — on the converted one. The API refuses a monetary value finer than two decimal places rather than rounding it, so both were rejected upstream and reached the buyer as a generic error. Now fixed at 2 at both points, via a named constant so it is visibly *not* the store's price precision.

Plain half-up rounding, deliberately. Sub-cent caps, away-from-zero rounding and zero-decimal currencies are all out of scope: the one value where the rounding direction would have mattered — an explicit cap of `0` — is refused by the grid instead.

### 4. Admin copy and the header padding

Help text now states that a cap of `0` is not allowed and, for `fixed_and_percentage` only, that the cap applies to the whole fee rather than the percentage alone. The four sites that described the cap as bounding only the percentage portion were already corrected in #404; nothing further was found.

Separately: WordPress core's `.widefat` pads row cells 10px horizontally but not header cells, so every column heading in the surcharge grid sat 10px left of the inputs beneath it. Matched.

### Tests

Every test verified by mutation:

| Mutation | Reddens |
| --- | --- |
| neuter the zero-cap grid check | `testSurchargeGridValidationNormalisesAndRejects` — "a cap of 0 must be rejected" |
| restore the `> 0` cap filter | `testBuyerFeeShareRelaysAConfiguredZeroCapVerbatim` |
| drop `round()` at payload assignment | `testBuyerFeeShareRoundsMonetaryValuesToTwoDecimalPlaces` |
| drop `round()` on the converted path | `testBuyerFeeShareCapRoundingToZeroRelaysZeroCap` |

```
$ php tests/unit/run.php
All tests passed.
$ make phpcs      exit 0 (0 errors)
$ make phpstan    [OK] No errors
```

### i18n

Three new strings, own append-only commit: `.pot` plus all three `.po`, `.mo` recompiled with `msgfmt`. `nl_NL` follows the surcharge-grid terminology already in that file.

Worth knowing: `nb_NO` and `sv_SE` have **no** surcharge-grid rows at all — they sit ~87 strings behind `nl_NL`, and that predates this change. These three are translated fresh rather than left silently English, but the wider gap in those two files is untouched and still owed.


dgjlindsay:
## Adversarial review — rounds 1-3

No summary had been posted on this PR before now; rounds 1 and 2 existed only in commit bodies. Recording the whole loop here.

### Round 3 findings (fixed in `13edf83`)

| # | Finding | Severity | Status |
|---|---|---|---|
| 1 | `assets/js/admin.js` `buildGridRow` read `stored[col] \|\| ""`. A stored **numeric** `0` is falsy in JS, so untick+retick rebuilt the row blank, the save dropped the cell, and the term's percentage relayed **uncapped**. The PHP renderer uses `isset()` and was correct — the two renderers disagreed. | **Blocker** | Fixed |
| 2 | Round 2's `is_numeric` gate on the cap **read** path turned a stored comma decimal (`"1,50"`) into absent, i.e. **uncapped**, where the pre-ticket code capped at 1.0. Failure direction was an overcharge. | High | Fixed — normalised, not dropped |
| 3 | The refusal was **invisible**. WooCommerce records a throwing field validator's message with `add_error` and nothing in core prints that bucket — `display_errors()` is the gateway's job and was never called. Typing a cap of `0` reverted the whole grid edit with no notice, on a page that looked like it saved. | High | Fixed |
| 4 | **Partial save.** Core skips only the field that threw, so refusing the grid still saved `surcharge_type`: `none → percentage` on a shop with a legacy stored zero went live at a cap of 0, which clamps the whole fee to nothing. | High | Fixed — both fields now refuse on one shared rule |

Finding 1 is the original overcharge this ticket exists to prevent, resurfacing through a path rounds 1 and 2 never examined.

### Closed as won't-fix (product decision, not an oversight)

The hidden-row lockout — a legacy stored zero on a term whose row is not rendered — is **accepted**. Also no migration for legacy stored zeros. Both deliberate.

### Verified correct, no change made

Re-checked at source rather than assumed: absent-vs-zero stays disambiguated on the read path; the zero-cap FX-gate exemption is applied at both the availability gate and the request builder; the `surcharge_type = none` cap-deletion regression introduced in round 1 and removed in round 2 is genuinely gone. Upstream, a cap of `0` **clamps the fee to zero** — it is not "uncapped" — and rejection of more than 2dp surfaces as HTTP 400.

### Choice worth flagging: comma-decimal caps

`"1,50"` is **normalised** on read, not treated as junk. It is this plugin's own accepted spelling — the grid validator and the merchant-minimum validator both apply the same `,`→`.` replacement before validating — so it reaches the stored option by hand edit, import, or an option written before that normalisation existed: the same arrival routes that justify relaying a stored zero at all. Normalising can only ever cap **lower** (a thousands separator reads as a decimal), which is the safe direction; dropping it relays an uncapped percentage. Genuine junk and negatives still read as absent.

### New test infrastructure

`assets/js/admin.js` had **no** test coverage, which is why a falsy-zero bug lived in it. `tests/js/admin-harness.js` assembles the real settings page and evaluates the real source, mirroring `wc-harness.js`.

One trap is worth naming, because it is the reason to trust the rest. On an already-complete document jQuery does not run a ready callback synchronously, and the readyList Deferred adds a further tick. A synchronous harness returns before `admin.js` has bound a single handler, every triggered event goes nowhere, and the **server-rendered** rows sit untouched in the DOM — which satisfies "the rebuilt row still holds its stored value" exactly. **Six tests passed that way before it was caught.** `loadAdmin` now drains the queue and `assertBootstrapped` fails loudly on a page the bootstrap never touched.

The `WC_Payment_Gateway` test stub also grows the `WC_Settings_API` save and error surface, because both the silent refusal and the partial save are properties *of* that machinery — a test calling one validator directly can see neither.

### Mutation verification

Every assertion was proven capable of failing. Source broken, suite confirmed **RED**, source restored, suite confirmed green:

| Mutation | Result |
|---|---|
| Restore `stored[col] \|\| ""` in `buildGridRow` | 3 tests fail |
| Print `"0"` for an absent cell (over-correct absence into zero) | 2 tests fail |
| Remove the ready-queue drain from the harness | `assertBootstrapped` fails, instead of the suite going green on untouched markup |
| Drop the comma normalisation from the cap read | *"a comma-decimal cap must cap at 1.50, not relay an uncapped percentage"* |
| Keep `is_numeric`, drop the negative guard | *"a negative stored cap must never be relayed"* |
| Remove `display_errors()` from `admin_options()` | *"the settings page must print the error notice"* |
| Remove the zero-cap gate from `validate_surcharge_type_field` | *"enabling must not survive a refused grid"* |
| Gate every `surcharge_type`, not only the percentage-bearing ones | the disabling-is-never-blocked test fails |

### Also in this commit

A code comment named a private service's schema class in this **public** repo. Reworded to describe the behaviour instead. It is pre-existing and already in public history, so it is flagged rather than treated as fixed — history was deliberately not rewritten.

### State

All 18 checks green. **Not merged**: one more adversarial round is owed on these round-3 changes before this can merge.

<sub>Review and summary written by Claude.</sub>

dgjlindsay:
Round 4 (post-rebase onto current staging): CLEAN. No new bugs or regressions introduced by conflict resolution; only pre-existing cosmetic/i18n findings unrelated to this PR remain, out of scope.

CI green, mergeStateStatus CLEAN.

— by Claude

PR #410, opened 2026-07-30
TWO-25292/ci(i18n): fail when a compiled .mo disagrees with its .po

dgjlindsay:
## Why

`.po` / `.pot` are text and auto-merge on rebase. The compiled `.mo` is binary and **cannot** merge — and resolving that conflict by keeping either side is silently wrong in one direction or the other:

- keeping **ours** resurrects a msgid the other side deleted
- keeping **theirs** drops a msgid this side added

WordPress reads only the `.mo`, so either way a translated shop ships wrong copy — or silently reverts to English — and nothing in CI could see it. This happened twice in one day: `staging` had removed a msgid (TWO-25287) while a feature branch added two (TWO-25283), and the resurrection was caught only because someone noticed and recompiled all three catalogues by hand.

## What

`.github/scripts/check-catalogues.sh`, wired as a `Gettext catalogues` job on the existing **Static analysis** workflow (no new workflow file). For every `languages/*.po`:

- recompiles it with `msgfmt` and compares against the committed sibling `.mo`
- fails on a `.po` with no `.mo`, a `.mo` with no `.po`, a `.po` msgfmt cannot compile, and a missing/malformed header
- locales are globbed, never listed

Failure output names the file, prints the exact `msgfmt -o <mo> <po>` to run, and states which side each diff line came from (`-` = live in the compiled catalogue but deleted from the `.po`; `+` = added to the `.po` but missing from the `.mo`).

## Decoded catalogues, not bytes

Byte comparison is simpler and every committed `.mo` here *is* currently byte-reproducible (verified with gettext 0.21). It was still rejected: the `.mo` encoder may change between gettext releases, the runners are self-hosted images this repo does not pin, and a byte check would then fail on every unrelated PR — worse than no check.

Instead both sides are decoded with the **same `msgunfmt` binary in the same job**, so whatever gettext the runner carries applies identically to both and cannot on its own produce a mismatch. That makes version pinning unnecessary rather than merely skipped; `msgfmt --version` is printed so a real toolchain change stays visible, and loss of byte-equality is reported as a note rather than a failure. Same reason fuzzy and obsolete entries need no handling — msgfmt drops them from both sides. Comment/line-reference churn in a `.po` correctly does *not* fail the gate, since it is not part of the compiled catalogue.

## Verified by mutation, not inspection

| | result |
|---|---|
| unmodified `staging`, all 3 locales | **PASS** |
| add a msgid to a `.po`, don't recompile | **FAIL** — diff shows `+msgid "…"` missing from the `.mo` |
| remove a msgid from a `.po`, don't recompile (the resurrection case) | **FAIL** — diff shows `-msgid "%s API Key"` still live in the `.mo` |
| then recompile the `.mo` properly | **PASS** |
| delete a `.mo` | **FAIL** — "no compiled catalogue for this source" |
| add a `.mo` with no `.po` | **FAIL** — "compiled catalogue has no .po source" |

An empty `languages/*.po` glob is also a hard failure, so the guard cannot pass vacuously.

## Follow-up worth a decision (not in this PR)

Whether the `.mo` files should be committed at all, or compiled at release time — that removes this class of conflict rather than guarding it. Not acted on; noted on TWO-25292.

Ticket: TWO-25292

### DIFF
--- a/.github/scripts/check-catalogues.sh
+++ b/.github/scripts/check-catalogues.sh
@@ -56,7 +56,7 @@ shopt -u nullglob
 
 echo "checking ${#po_files[@]} catalogue(s)"
 
-tmp=$(mktemp -d)
+tmp=$(mktemp -d) || fail "could not create a temporary directory"
 trap 'rm -rf "$tmp"' EXIT
 
 status=0
@@ -97,8 +97,22 @@ for po in "${po_files[@]}"; do
         continue
     fi
 
-    if diff -u --label "$mo (committed)" --label "$po (recompiled)" \
-        "$tmp/committed.po" "$tmp/generated.po" >"$tmp/diff.txt"; then
+    # diff exits 0 for identical, 1 for differing, 2 for trouble. Folding 2
+    # into 1 would report a broken diff as a catalogue mismatch and send the
+    # developer to recompile a catalogue that was fine.
+    diff_status=0
+    diff -u --label "$mo (committed)" --label "$po (recompiled)" \
+        "$tmp/committed.po" "$tmp/generated.po" >"$tmp/diff.txt" 2>"$tmp/diff.err" ||
+        diff_status=$?
+
+    if [ "$diff_status" -gt 1 ]; then
+        echo "::error file=$mo::could not diff this catalogue against its source"
+        sed 's/^/  /' "$tmp/diff.err"
+        status=1
+        continue
+    fi
+
+    if [ "$diff_status" -eq 0 ]; then
         # Byte equality is not required, but its loss is the only warning that
         # the runner's gettext has started encoding differently, so say so.
         if cmp -s "$mo" "$tmp/generated.mo"; then
--- a/.github/scripts/check-catalogues.sh
+++ b/.github/scripts/check-catalogues.sh
@@ -1,0 +1,138 @@
+#!/usr/bin/env bash
+#
+# Every committed .mo must agree with its sibling .po (TWO-25292).
+#
+# WordPress reads only the compiled .mo, so a .mo that has drifted from its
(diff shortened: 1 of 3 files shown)


dgjlindsay:
## Self-review summary — 2 rounds to clean

Repo does not enforce a separate reviewer (`reviewDecision: null`), so this went through the self-review lane rather than merging on green CI alone.

**Round 1 — 3 findings, all medium:**

1. `diff` exits `2` for trouble, not only `1` for differing — folding the two together would report a broken diff as a stale catalogue and send someone to recompile one that was fine. **Fixed:** `diff_status` is now tested three ways (0 / 1 / >1), and >1 reports "could not diff this catalogue" with diff's own stderr.
2. `mktemp -d` failure was only caught implicitly by `set -e`, with no message. **Fixed:** named explicitly.
3. `sudo apt-get` assumes a sudo-capable Debian-family runner. **Disputed, not changed** — the runner was verified to be Ubuntu jammy and the install step's own failure is the error signal; a platform guard would add a second thing to keep in sync with no new information.

**Round 2 — 0 findings.** Control flow re-checked after the round-1 restructure: every failure path still sets `status=1`, the summary still fires, and no branch became unreachable. Clean round.

**Mutation re-run after the fixes** (the guard has to be able to fail, so this was re-proven, not assumed): unmodified tree PASS; msgid added to a `.po` without recompiling FAIL; msgid removed from a `.po` without recompiling FAIL; recompiled FAIL→PASS.

**Also checked:** the guard was dry-run against the head of the other open PR that touches catalogues — it passes there, so this does not land as a surprise red build on in-flight work.

<sub>Review by Claude.</sub>

PR #411, opened 2026-07-30
docs: replace cross-repo references with behaviour descriptions

dgjlindsay:
## What

Five places named an internal repository rather than describing behaviour:

- `class/WC_Twoinc.php` — minimum-order docblock
- `class/WC_Twoinc_Payment_Terms.php` — pricing-request comment
- `assets/js/twoinc.js` — a `@TODO` comment
- `.env.example` — the local-dev API base URL comment
- `tests/e2e/` — one test helper was named after an internal repository; renamed to `two-api.ts`, which is also the more accurate name for what it wraps

## Why

This is a public repository. Prose, comments and filenames here should describe behaviour against the documented contract, not name internal repositories.

## Risk

Comment and docs text, plus one test-helper rename with its two importers updated. No plugin behaviour change.

Verified: `php -l` clean on both PHP files, `node --check` clean on the JS, and `tsc --noEmit` clean in `tests/e2e` after the rename. No workflow, Makefile or other file references the old filename.

Part of item #30.10. Kept deliberately minimal in `class/WC_Twoinc.php` (one comment line) to avoid conflicting with in-flight branches.

## Not changed

Cross-references to the public sibling plugin repositories (stylesheet comments, `tests/js/*`, `tests/unit/run.php`) are left alone by decision — those repositories are public.


dgjlindsay:
## Adversarial review — one round, no blocking findings

Scope: five prose substitutions (env example, one JS TODO comment, two PHP block comments) plus one test-helper file rename with two importers updated. No behavioural line changes.

**Checked and clean**

- **Rename is complete.** The helper module was renamed and both importing specs updated. Grepped the entire head ref for the old basename — in imports, workflow files, `Makefile`, `package.json`, Playwright config and docs alike: **zero occurrences remain**. The new basename is unique in the tree (no pre-existing file was clobbered; the rename is 100% similarity, content untouched). The browser-test job under the `Tests` workflow, which actually resolves those two imports at runtime, is green — that is the real proof the rename holds, not the grep.
- **Behaviour still described correctly.** Each substituted comment keeps its point: the minimum-order value is the platform default with merchant override and is enforced server-side at order create/intent; the surcharge request field is required by the pricing request and is hardcoded for cross-plugin parity. No semantics were softened or inverted.
- **No load-bearing comment lost.** All five are in-place substitutions of a service name inside an otherwise-intact sentence. The parity rationale for the hardcoded recourse flag, and the note that neither plugin has admin config for it yet, both survive.
- **No foreign source text left in the touched files.** Grepped the head ref for schema names, source expressions and test identifiers: zero hits.
- CI green: unit tests PHP 7.4-8.4, Jest, phpcs, phpstan, gettext catalogues, pre-commit, browser tests, upgrade smoke, boot smoke on WooCommerce 10.8.1 and 10.9.4, version-bump gate.

**Non-blocking nit, not worth a push**

The env-example line now reads "a locally running API behind portal.localhost", which is a little vague where it used to name the component. It is accurate and the URL directly beneath it resolves any ambiguity, so this is style, not substance. Left alone rather than reopening a green PR.

Verdict: merge.

---
<sub>Review written by Claude.</sub>

PR #412, opened 2026-07-30
feat(TWO-25288): plugin-owned company-search hints for the empty and too-short field

dgjlindsay:
TWO-25288, elements 3 and 4: the two hints the company-search field owes the buyer before it can search anything.

## Element 3 — empty-field hint

The empty field showed nothing, so there was no way to tell the control was a search box rather than a dropdown listing nothing. It now carries a translatable `Enter company name to search` placeholder.

The pay-for-order page builds its own copy of the company select, and its empty `<option>` had no `value` attribute — so it took its own non-breaking-space label as its value, the widget read that as a selection, and the placeholder never painted. It carries `value=""` now, matching what the checkout page renders through WooCommerce's field API.

## Element 4 — min-chars hint

This was WooCommerce core's own copy, which counts down the characters still *missing*. The same field therefore said "3 or more" before any keystroke and "1 or more" after two — describing a moving target rather than the requirement. It is now a plugin-owned translatable string stating the threshold itself, and the borrow of the core strings is gone.

## One threshold, one place

`twoincSelectWooHelper.companySearchMinLength` in `assets/js/twoinc.js` is now the only place the number appears in the company-search path. The widget's `minimumInputLength` reads it, the "not in the list" button's visibility rule reads it, and the hint is interpolated from it.

That is why the PHP string keeps an unresolved `%d`: PHP registers the template, the browser fills in the number it actually enforces. There is no second copy of the threshold that could claim one minimum while the widget enforces another. A test asserts PHP has *not* resolved it.

## Translations

Both source strings added to the template and translated for nb_NO, nl_NL and sv_SE. Each `.mo` was recompiled with `msgfmt` from its merged `.po`, so the catalogue check agrees. The min-chars entry is marked `php-format`, so `msgfmt` now refuses a translation that drops the `%d` — without it the buyer would be shown no number at all.

## Tests

92 pass. Every new test was mutation-verified: the source was broken, the test confirmed RED, and the source restored. Eight mutations, each failing at least one test —

- `minimumInputLength` back to a literal `3`
- `inputTooShort` back to core's countdown strings
- the `placeholder` param deleted
- the hint's `%d` filled from a literal instead of the constant
- the pay-for-order empty option's `value=""` removed
- the harness's empty option's `value=""` removed (proves the DOM assertions bite)
- the PHP localisation key renamed
- the `%d` resolved in PHP

Coverage includes the real widget attached to the real option markup, so the non-breaking-space label cannot silently defeat the placeholder, plus the inverse case; and the localisation keys are pinned on both sides, since a renamed key would otherwise leave the browser serving untranslated English with the rest of the suite green.

Out of scope, untouched: the payment-tile company field and the organisation-number field.


dgjlindsay:
## Adversarial review — one round, then a clean round

Three lenses (correctness/edge cases, platform idiom + i18n, tests), every finding verified against the fetched ref before acting — and one of them reproduced against real checkout markup rather than argued about.

**The serious one: the hint was being submitted as the buyer's company name.**

The checkout-input snapshot reads each select2 container's text and only treats it as empty when it finds a text-node child. The old empty option rendered as a non-breaking space — a text node that trimmed to empty. The new placeholder renders as an *element*, so no text node exists, and the snapshot captured the hint copy itself. From there it flowed into the company-name getter, poisoned the cached company data on first page load, and was written into the billing-company field **that is posted with the order**. It also reached the sole-trader signup prefill and the intent-approved sentence. Fixed at the source: a container rendering a placeholder is by definition unselected, so it snapshots as empty, and the placeholder is excluded from the restored sub-nodes — which also fixes the hint rendering twice after a restore from storage.

**Why CI could not see it, which is the part worth keeping:** the test harness emitted a checkout form without the attribute WooCommerce's real form carries and that the snapshot function looks the form up by — so every test silently exercised the no-form early return. Fixed. And *independently*, no test called the snapshot, restore or company-name paths at all. Two separate reasons for the same blind spot; both closed, nothing papered over. The suite still passes with the harness corrected.

**Also acted on:**
- The threshold interpolation matched only the bare placeholder, not gettext's positional form a translator is entitled to use — that would have shown the raw token. Now matches both. The msgid is unchanged.
- Added a translators comment, and dropped the format flag that claimed PHP resolves the string when in fact JS does.
- One negative assertion would still have passed if PHP had resolved the placeholder — the exact drift it claimed to prevent. Now asserts the emitted value positively.

**Mutation-verified**, not asserted: with only the snapshot fix reverted, five tests fail, including the one that matters —

```
● … › is not written into the company field posted with the order
    Expected: ""
    Received: "Enter company name to search"
```

and the doubled-hint case reproduces as `Expected 1, Received 2`. Restored, 99 tests green.

**A latent harness hazard found while doing this, worth knowing:** the posted-field test needs *real* timers. jQuery runs ready callbacks on a macrotask, so installing fake timers before yielding freezes the plugin's page-load bootstrap before it registers anything — the first version of that test passed under mutation for precisely that reason, i.e. it was silently vacuous. Documented in the test.

**Catalogues:** only comment and flag lines changed — no msgid, msgstr or header. Those are not carried in the compiled catalogues, so no recompilation was needed and no compiled catalogue was touched. Verified with the repo's own catalogue check: all catalogues agree with their sources, and the CI catalogue job is green.

Second round over the fixes found nothing new. Full matrix green including e2e, both boot smokes and the upgrade smoke.

Not verified in a browser.

<sub>Review and this comment written by Claude.</sub>

PR #413, opened 2026-07-30
feat(TWO-25288): make the manual-entry affordance a keyboard-reachable dropdown option

dgjlindsay:
TWO-25288 element 5. Makes `My company is not on the list` a real, keyboard-reachable row inside the company-search dropdown, and makes the way back out of manual entry reachable too.

## What was wrong

The affordance was a hidden `<div>` in the checkout billing form, cloned into the dropdown and appended to the results list's **parent** — outside the subtree the search field's `aria-owns` points at. No `role`, no `id`, no payload, no `tabindex`. It was not reachable by keyboard at all. The link back out of manual entry was an unreachable `<div>` as well.

Two further defects in the same block:

- the handler that showed it was bound **inside a polling callback** that fired long after the dropdown opened, so a fast typist's first keystrokes were dropped;
- it was re-bound on **every dropdown open with no matching `.off()`**, accumulating one duplicate handler per open.

## What this does

**A genuine pseudo-option, last row inside the list.** It mirrors, attribute for attribute, what the picker's own option factory produces for a real result row: `role="option"`, `data-selected="false"`, `tabindex="-1"`, a stable id, and a payload whose `_resultId` matches that id. Asserted by comparing against a row the widget itself builds, rather than against a hand-written list that could drift out of step with the library.

`tabindex="-1"` is not decoration. On every arrow keypress the picker calls `.focus()` on the highlighted row — its own source says that is required for screen readers to work — and an `<li>` with no tabindex cannot take focus, so that call is a silent no-op and focus stays stranded on the *previous* row while this one is visually highlighted. `_resultId` is what the picker copies into `aria-activedescendant`, on both the dropdown's search field and the selection container. The `--selectable` class is belt-and-braces only: that string does not occur anywhere in the bundle pinned here, so it is inert today and is described as such in the source — it costs nothing and keeps the row navigable if a host platform ever swaps the picker for one that keys off a class.

> **Why `data-selected` is present here and absent on the sibling platforms.** The other implementations of this same element under TWO-25288 **drop** that attribute. That is not an inconsistency to reconcile, and neither side is a bug.
>
> The attribute is not part of any published ARIA contract. It is an implementation detail of the specific picker build a platform ships, and the platforms ship different ones. The picker this plugin uses comes from the host platform rather than from this repo, and it is an older lineage: verified against the dependency actually installed here, it derives its entire navigable result set from the presence of that attribute, and it treats the already-selected value on a row as a signal to **close** the dropdown rather than to select the row. So omitting the attribute makes the row unreachable, and setting it to the already-selected value makes the row look reachable while doing nothing — both failure modes are pinned by mutations here.
>
> A platform whose picker derives its navigable set from a CSS class instead gets nothing from the attribute, and is right to leave it out. Read the two as the same rule applied to two libraries. The mirror of this note is in the sibling change; see TWO-25288 for the per-platform verification.

**One activation path for keyboard and mouse.** The picker turns both Enter on the highlighted row and a click on it into the same internal select, and fires a preventable pre-event first. Intercepting that one event covers both, and the previous click-only handler is gone — keeping it would have double-fired on every mouse activation and let the two inputs drift apart. Prevention means nothing is written: no company name, no company id, no approval call, no address lookup.

**Defect 1 fixed** by delegating the visibility handler on `<body>`, so it exists before the buyer's first keystroke rather than after a poll notices the field appeared. **Defect 2 fixed** by namespacing it and calling `.off()` before `.on()`, so the repeated re-binds (the timed re-run of the search setup, and every return out of manual entry) leave exactly one.

**Timing unchanged from the reference:** the row appears on `input` once the term reaches the threshold, *before* the debounced request goes out. No "has a search run" gate — a buyer who already knows their company is not in the registry should not wait for a round trip to find that out. The threshold is read from the one shared constant introduced earlier in this ticket, not re-hard-coded.

**Re-appended after every render.** The picker empties the whole list on each new result set, each message, and the searching state; a MutationObserver on the list covers all of them without guessing at internal events.

**The way back out is a `<button type="button">`** rather than the old `<div>`, so it is focusable and Enter/Space-activatable with no keydown bridge of our own.

## Found by review, fixed here

**The row died permanently after a widget re-create.** Clearing the selected company — the × on the floating company id, i.e. the "un-pick that, let me search again" gesture that most often precedes needing this row — rebuilds the picker and so the results list. The render watcher was installed from the bind path and keyed on the old list, so it was never re-installed: the row would appear on a keystroke and then be wiped by the first render, for the rest of the page. It installs from the per-keystroke sync now, so any re-creation heals itself without a third bind site to keep in step, and at most one watcher is live with the previous disconnected.

**Focus is no longer dropped on either mode switch.** Entering manual entry destroyed the widget and focused nothing, leaving `activeElement` on `<body>`; leaving it hid the button that had focus. Both paths now move focus, and the helper that does it reports failure rather than claiming success when the field is absent.

**Entering manual entry clears the real company field too**, not only the display one — otherwise the manual field is pre-filled with the company the buyer has just said is not theirs, next to an empty number field. The exit path already cleared that mirror, so this was asymmetric as well as wrong.

**A manual-entry → sole-trader → business round trip no longer strands the buyer.** That path snapshots the current search setting, which manual entry had written as disabled, and restores it — so the re-enable early-returns and the link back to search stayed hidden with no route to the picker at all.

**A repeated activation in one tick switches once.** The prevented selection leaves the dropdown open, so the row was still there to be hit again; it is removed up front.

## The pay-for-order surface

That page renders its own copy of the company inputs and runs the same search binding, but it does not render the billing-form view — so the clone source was an empty set there and the affordance was silently absent, with no way into manual entry and no way back.

Both strings moved to the localised text map and both nodes are built in JS, so that page gets the affordance with no template of its own. Building in JS rather than adding markup to the second view, for two reasons: the row has to be an `<li>` pseudo-option, which a cloned `<div>` cannot become, so the node was going to be constructed either way; and one string source beats two.

That exposed a latent trap on the same page: it lays each company input out in a wrapper carrying its own hidden state, independent of the field inside it. Revealing the manual field alone would have left manual entry with nowhere to type. Wrapper visibility now follows the field's, which is a no-op on the checkout page. This was already reachable via the merchant setting that disables company search; it just had no buyer-facing route until now.

## Catalogues

The two msgids move from the view to the text map, so their reference comments move with them. Msgids and every translation are unchanged, and the compiled catalogues are byte-identical — the gettext gate confirms it.

## Verification

`npm run test:js` — **138 passing across 4 suites**, of which 42 are this element's own. The new suite `tests/js/company-search-manual-entry.test.js` drives the real widget rather than a mock, because the row's reachability *is* a property of the widget's own navigation code — a mock would have to reproduce that code correctly to catch a regression in it, which is exactly the assumption that let the previous unreachable implementation ship.

Reachability and announcement are asserted by driving the picker's own navigation and focus routines and then reading `document.activeElement` and all three `aria-activedescendant` sites, not by inspecting markup and hoping.

### Applied-proof mutation pass

Every mutation was re-run under a shared applied-proof harness that refuses to report a result it cannot trust: the tree must be clean before mutating, `git diff` must be non-empty after applying or no test result is printed at all, and the file is restored and re-checked clean on every exit path. Anchors are matched as plain text, never regex.

The harness was copied into a private scratchpad and md5-verified there rather than run from a shared path — a shared mutable script is as dangerous as a shared mutable file. Run against md5 `2955b8b0c29851248d5f85a7c2e2720d`, byte-identical to the published snapshot, HEAD confirmed unmoved across the pass, tree clean after.

**36 mutations re-run · 36 applied and RED · 0 survived · 0 did-not-apply · 0 aborts.** Every verdict came from `npm run test:js`; no PHP or shell runner was ever put through the harness, so none of these verdicts depends on runner-shaped output parsing.

Anchors were validated up front against the current source and every one had to resolve to **exactly one** occurrence before anything ran. That gate earned its keep twice: one anchor was ambiguous because the same assignment appears in both the manual-entry path and the sole-trader suppression path, and mutating whichever came first would have tested something else entirely; and the exact-count-1 rule also covers indentation drift by construction, since a mis-indented literal resolves to zero matches and is reported stale rather than silently skipped.

Notable kills: appending to the list's parent instead of the list (19 tests); `data-selected="true"` (attributes plus every activation test); dropping `tabindex` (three, including the focus assertion); dropping `_resultId` (payload plus announcement); a `hasSearched` gate (19); hard-coding the threshold; removing the render watcher; installing it only from the bind path; never disconnecting the previous one; dropping the re-entrancy guard; dropping `preventDefault`; intercepting every activation rather than only the sentinel; and `focusVisibleCompanyField` claiming success unconditionally.

### Four survivors, and what was done about them

The pass found four mutations that left the suite green. None was a bug; three were code claiming a guarantee nothing checked, and they were **removed** rather than given a test written to fit them.

- **A same-tick activation latch was unreachable.** Removing the row before deferring already prevents a repeated activation, so no path reached the latch. Deleted.
- **The bind function's widget argument was not load-bearing.** Reassigning it to a fresh lookup of the same field changed nothing and no test noticed — everything in the body names the company-search field, so the parameter could only ever be that one element. Signature dropped.
- **An eager render-watcher install was redundant.** The row cannot exist before the buyer has typed to the threshold, and typing runs the sync, which installs the watcher. One install point now, which is also what makes a re-created widget heal itself.
- **The one-watcher-per-node guard needed more than one keystroke to be visible.** The only genuine test gap of the four: with a single install point, one keystroke cannot distinguish "installed once" from "re-installed per character". The test now types four times, and the mutation goes red.

### Vacuity audit

Audited for two distinct classes of assertion that cannot fail.

The **structural-token** class — a comparison against an opaque empty-object token, which matches any other empty-object token — is absent outright: this suite uses no test doubles, so there is no `toHaveBeenCalledWith` and nothing to spy on.

The **already-empty collection** class was present twice, and the detection method was reading *which* test caught each mutation rather than trusting the red.

- `no company value is written` rested on three "nothing happened" checks that a recorder which was never wired up satisfies. Making the interception unconditional left it green. It now selects an ordinary row the ordinary way first and requires that to be recorded, and that mutation goes red.
- `absent below the threshold` was genuinely unfalsifiable — the row had never been in the list, so no mutation could turn it red, and none of the original 25 did. It now reaches the threshold first and asserts the row is present before asserting it goes away.

One further vacuity was caught in a *test* rather than the source: the observer count included the MutationObserver the widget constructs for itself, so it read 2 where the plugin made 1 and would have failed for a reason unrelated to this code. It is keyed on the observed node now.

Two harness traps were hit and fixed rather than worked around: `:hidden` matches every element under jsdom (zero dimensions), so the visibility assertions read the inline `display` this code actually sets; and the first reachability test asserted only the selection container's `aria-activedescendant`, which the dropped-`_resultId` mutation survived — the dropdown's search field is the element the field's `aria-owns` points from, and all three are asserted now.

## Not verified — needs a real browser

- That the row *looks* like an option row. It inherits the picker's option padding and highlight instead of a hand-rolled margin. The highlighted-row text colour is no longer left to the cascade — review found the id rule outranked the picker's highlight rule, leaving brand blue on a blue background at roughly 2.5:1, and it is now restated at matching specificity for both theme prefixes. jsdom resolves no box geometry, so the rendered result is still unobserved.
- That the `<button>` reset renders identically to the old `<div>` link.
- Screen-reader announcement. The ARIA wiring is asserted at the attribute level and driven through the widget's own navigation, which is as far as jsdom reaches.
- The pay-for-order page end to end. Its wrapper behaviour is unit-covered and its markup verified against the fetched ref, but the page itself was not loaded.

Not merged — the coordinator runs the adversarial review.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -167,19 +167,35 @@
   white-space: nowrap;
 }
 
+/*
+ * Manual-entry row inside the company-search results list (TWO-25288).
+ *
+ * It is an option row now rather than a floating div beside the list, so the
+ * hand-rolled margin is gone: the row inherits the picker's own option padding
+ * and highlight, which is what keeps it in step with the company rows above it
+ * when arrowed onto. Kept as a flat single-id selector so brand overlays that
+ * recolour it keep matching without raising their own specificity.
+ */
 #company_not_in_btn {
   cursor: pointer;
   color: #3043d1;
-  margin: 4px 12px;
 }
 
+/*
+ * The link back out of manual entry. A <button> now, for keyboard
+ * reachability, so the browser's button chrome has to be reset for it to look
+ * like the link it did before.
+ */
 #search_company_btn {
   text-align: right;
   cursor: pointer;
-  padding-top: 5px;
+  padding: 5px 0 0;
   color: #3043d1;
   position: absolute;
   right: 4px;
+  background: none;
+  border: 0;
+  font: inherit;
 }
 
 /* Payment terms chip selector (TWO-24751) */
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -182,6 +182,24 @@
 }
 
 /*
+ * Highlighted state (TWO-25288). NOT cosmetic — a contrast fix.
+ *
+ * The picker paints its highlighted row white-on-blue, but its own rule is a
+ * class selector (specificity 0,3,1) and the id rule above is 1,0,0, so the
+ * brand blue above wins on the text while the blue background still applies:
+ * roughly 2.5:1, under the 4.5:1 floor, where every real company row in the
+ * same list goes white-on-blue. Restated at the same id specificity so this row
+ * matches its neighbours instead of going unreadable the moment it is arrowed
+ * onto. Both theme prefixes, because which one is present depends on the host
+ * theme's picker build.
+ */
+#company_not_in_btn.select2-results__option--highlighted,
+.select2-container--default #company_not_in_btn.select2-results__option--highlighted,
+.select2-container--classic #company_not_in_btn.select2-results__option--highlighted {
+  color: #ffffff;
+}
+
+/*
  * The link back out of manual entry. A <button> now, for keyboard
  * reachability, so the browser's button chrome has to be reset for it to look
  * like the link it did before.
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -98,9 +98,14 @@ let twoincSelectWooHelper = {
    *
    * It has to be an id and not a flag: the picker stringifies every navigable
    * row's `data.id` when it recomputes selected state, so a row without one
-   * throws and takes the whole results render with it. And it must not be
-   * mistakable for a company: `processResults` below maps each hit's id to the
-   * company NAME, so the sentinel is deliberately not a plausible one.
+   * throws and takes the whole results render with it.
+   *
+   * `processResults` below maps each hit's id to the company NAME, so this
+   * value shares a namespace with company names and the underscores are there
+   * to keep it out of the way. That is a convention, NOT an enforced
+   * invariant — nothing rejects a registry name that happened to equal it, and
+   * no test claims otherwise. If that ever needs to be a guarantee, the
+   * discriminator has to move off `id` onto a field of its own.
    */
   manualEntrySentinelId: "__twoinc_manual_entry__",
 
@@ -186,20 +191,30 @@ let twoincSelectWooHelper = {
    * genuine option row now, which is what makes it arrow-reachable and
    * announced without a bespoke keydown bridge.
    *
-   * Every attribute here is load-bearing:
+   * This mirrors, attribute for attribute, what the picker's own option
+   * factory produces for a real result row. Three of them are load-bearing:
    *
-   *  - the bundled picker's navigable set is the rows carrying a
-   *    `data-selected` attribute, while select2 4.1 keys off the
-   *    `--selectable` class instead. A host theme may supply either, so the
-   *    row carries both and is reachable under both.
-   *  - `data-selected` must stay "false". "true" means "already picked", and
-   *    the picker routes activation of an already-picked row to closing the
-   *    dropdown instead of selecting it — the row would look reachable and do
-   *    nothing.
+   *  - `data-selected` is how the picker derives its navigable set, and it must
+   *    stay "false". "true" means "already picked", and the picker routes
+   *    activation of an already-picked row to closing the dropdown instead of
+   *    selecting it — the row would look reachable and do nothing.
+   *  - `tabindex="-1"` makes the row programmatically focusable. On every arrow
+   *    keypress the picker calls `.focus()` on the highlighted row, and its own
+   *    source says that is required for screen readers to work. An <li> with no
+   *    tabindex cannot take focus, so that call is a silent no-op and focus
+   *    stays stranded on the PREVIOUS row while this one is visually
+   *    highlighted — the reader keeps announcing the wrong thing.
    *  - the payload goes on with jQuery `data`, which is where the picker reads
    *    each row's data from, and it must carry `_resultId` matching the li's
-   *    id: that is what the picker copies into aria-activedescendant when the
-   *    row is highlighted. Without it the row is reachable but silent.
+   *    id: that is what the picker copies into `aria-activedescendant`, on both
+   *    the dropdown's search field and the selection container. Without it the
+   *    row is reachable but silent.
+   *
+   * `select2-results__option--selectable` is belt-and-braces only. The bundle
+   * pinned here never reads it — it is a select2 4.1 concept, and that string
+   * does not occur anywhere in this build. It costs nothing and keeps the row
+   * navigable if a host platform ever swaps the picker for one that keys off
+   * the class instead of the attribute. Do not describe it as required.
    *
    * @returns {Object} jQuery-wrapped <li>
    */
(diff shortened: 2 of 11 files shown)


dgjlindsay:
## Independent review summary — element 5 findings fixed

Four fresh, independent review rounds against the fixes for the B1/B2/B3/B5 findings from the earlier element-5 audit. Head at this comment: `9c5f67d`.

**Round 1** (against `e003067`) — B1 fixed (clear the disowned company's registry address on manual entry), B2/B3/B5 test coverage added. Found: the address clear fired unconditionally, wiping a buyer's own account-prefilled address on the ordinary no-pick path.

**Round 2** (against `5b7f02f`) — fixed by gating on whether a pick had happened. Found the gate signal itself (`#company_id` non-empty) was wrong in both directions: false-positive for account-restored/sole-trader buyers who never picked, false-negative for a picked company with no organisation number whose lookup still ran.

**Round 3** (against `60588e9`) — replaced the gate with an explicit `registryAddressApplied` flag, set only where a registry lookup actually writes address data, reset on both consuming paths. Found one coverage gap: `clearSelectedCompany`'s flag reset was untested (deleting it stayed green). Also surfaced a pre-existing, out-of-scope timing issue (an in-flight lookup that resolves after the buyer has already moved on) — not introduced by this PR, not fixed here, flagged separately for a follow-up ticket.

**Round 4** (against `9c5f67d`, test-only) — CLEAN, round converged. No new findings.

Every fix and every closed test gap was confirmed by mutation testing (apply a real code mutation, prove the suite goes red on the specific test that claims the coverage, revert). 148/148 tests pass, 18/18 CI checks green.

— reviewed and fixed by Claude

PR #414, opened 2026-07-30
feat(TWO-25288): open the search dropdown when the buyer asks to search again

dgjlindsay:
## What

Returning from manual company entry to company search now opens the search dropdown and puts the
caret in its search field, in the same gesture as the click.

## Why

`exitManualCompanyEntry` re-attached the picker and focused its **closed** combobox, so the buyer
who had just asked to search again was handed a closed control and had to click a second time
before there was a search box to type into.

The explicit `focus()` is not redundant with the picker's own. The picker focuses its search field
from a listener on its `open` event, and this plugin already carries a polling focus fix
(`waitToFocus`, wired to `select2:open`) precisely because that focus does not reliably land — a
poll that can take ~2.4s to get there. Focusing here makes the caret's arrival synchronous with
the click; the poll then finds the field already focused and no-ops.

The previous combobox-focus chain is kept as the fallback for a surface with no picker attached
(the pay-for-order page renders a different field set), which is what the new helper's boolean
return is for.

## Ordering

The open happens **after** `toggleBusinessFields()`. Opening positions the dropdown against its
container, and that container is only laid out once the business fields have been shown.

## Tests

Six tests in `tests/js/company-search-manual-entry.test.js`, on the real widget, asserting the
picker's own rendered state rather than a spy:

- the container carries the open class and the results list exists;
- `document.activeElement` **is** the dropdown's search field, and is neither the closed combobox
  nor `<body>`;
- end to end — type into whatever holds focus with no second click, and the manual-entry row
  reappears;
- opening an already-open dropdown is idempotent (one dropdown, focus retained);
- no picker attached, and no company select at all, both report failure rather than lying.

Both halves are pinned independently, verified by mutation: removing the open call fails 3 tests,
and removing only the explicit focus (leaving the open) fails 3 tests.

```
$ npm run test:js
Test Suites: 4 passed, 4 total
Tests:       154 passed, 154 total
```

Part of TWO-25288.

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -1199,20 +1199,27 @@ let twoincDomHelper = {
     // After toggleBusinessFields, deliberately. Opening the dropdown positions
     // it against its container, and that container is only laid out once the
     // business fields have been shown.
-    if (twoincDomHelper.openCompanySearchDropdown()) return;
-
-    // Fallback for a surface with no picker attached (the pay-for-order page
-    // renders a different set of fields). Mirrors the enter path: the button
-    // that had focus is now hidden, so without this focus is stranded on a
-    // display:none element.
-    //
-    // NOT #billing_company_display — the picker hides that <select> and moves
-    // its accessible role onto the rendered combobox, which is the element
-    // carrying tabindex and the one a buyer can actually see.
-    if (
-      !twoincDomHelper.focusVisibleCompanyField("#billing_company_display_field .select2-selection")
-    ) {
-      twoincDomHelper.focusVisibleCompanyField("#billing_company_display");
+    if (!twoincDomHelper.openCompanySearchDropdown()) {
+      // Fallback for a surface with no picker attached (the pay-for-order page
+      // renders a different set of fields). Mirrors the enter path: the button
+      // that had focus is now hidden, so without this focus is stranded on a
+      // display:none element.
+      //
+      // Reached ONLY when no dropdown was opened. Running it alongside an open
+      // dropdown would park focus on the collapsed combobox while the picker is
+      // expanded behind it — a worse state than either outcome on its own,
+      // because the buyer's keystrokes would go nowhere the open list can see.
+      //
+      // NOT #billing_company_display — the picker hides that <select> and moves
+      // its accessible role onto the rendered combobox, which is the element
+      // carrying tabindex and the one a buyer can actually see.
+      if (
+        !twoincDomHelper.focusVisibleCompanyField(
+          "#billing_company_display_field .select2-selection"
+        )
+      ) {
+        twoincDomHelper.focusVisibleCompanyField("#billing_company_display");
+      }
     }
   },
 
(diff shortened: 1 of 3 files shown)


dgjlindsay:
Adversarial self-review round 1 — one real finding, fixed in 577c896.

**Mixed focus state (fixed).** `openCompanySearchDropdown` reported whether *focus* landed, and
`exitManualCompanyEntry` read that as "did anything happen". A dropdown that opened but whose focus
call failed therefore fell through to the fallback chain and parked focus on the collapsed combobox
in front of the expanded list — worse than either outcome alone, because the buyer's keystrokes go
nowhere the open list can see. The helper now reports whether the *dropdown* was opened, which is
the question the caller is actually asking, and the fallback is unreachable while a dropdown is
open. Pinned by a test that forces the focus call to fail and asserts the combobox does **not** take
focus; reverting the fix fails exactly that test.

Considered and cleared:

- **Immediate re-close by the picker's own outside-click handler.** It is bound on `mousedown`, and
  the mousedown of the activating click has already happened by the time the `click` handler opens
  the dropdown, so there is no later mousedown to close it. Keyboard activation produces no
  mousedown at all.
- **Focus race against the dropdown rebuild.** The search field exists synchronously after
  `select2("open")` — the tests focus and then type into it in the same tick — and it is looked up
  after the open, never cached, so it is the node this open created.
- **Focus fight with the existing `waitToFocus` poll.** The poll targets the same input and
  no-ops when it is already focused. It is reached through the pre-existing `select2:open`
  binding, identical to a buyer-initiated open, so this introduces no new path.
- **Observer/listener accumulation.** Nothing new is bound here. The affordance's observer is
  installed from the per-keystroke sync and keeps its at-most-one invariant, covered by the
  existing "one handler and one observer, across every re-bind" tests.
- **Pay-for-order surface.** No picker attached, so the helper returns false and the previous
  combobox-focus chain runs unchanged — tested both for the missing picker and for no company
  select at all.
- **Accessibility.** Both the expansion and the focus move are the direct result of the buyer
  activating a control, so WCAG 3.2.1 is not in play; `aria-expanded` is the picker's own.

No further findings. Not merging — leaving this for a human look.

<sub>Review by Claude.</sub>

dgjlindsay:
### Adversarial review round 2 — clean

Confirming round after round 1 found and fixed the mixed focus/dropdown-opened state bug (sha 577c896). Verified on the current head:

- Round 1's fix (`openCompanySearchDropdown` reporting dropdown-opened rather than focus-succeeded) has full negative-path test coverage (mixed-state, no-picker, no-select-at-all cases).
- No new logic bugs: `select2("open")` is idempotent, mousedown-vs-click ordering rules out re-close by the outside-click handler, and the explicit focus call doesn't race the existing `waitToFocus` poll (same target, no-ops if already focused).
- Test suite exercises real behaviour, not vacuous assertions.
- `npm run test:js`: 155/155 pass across 4 suites.
- No accessibility or scope-creep findings.

No new findings this round. Merge-eligible.

*Review performed by Claude.*

PR #416, opened 2026-07-31
fix: company-search manual-entry link, focus trap, and placeholder styling

dgjlindsay:
## Summary

Four buyer-reported bugs in the checkout company-search widget's "my company is not on the list" affordance and dropdown, found via live testing:

- **Reachable only by scrolling / arrowing through every result.** The affordance was a pseudo-option `<li>` living inside `.select2-results__options` (the scrollable results list), so it was only visible after scrolling past however many results came back, and the only way to reach it by keyboard was arrowing down through every result — it deliberately carried `tabindex="-1"`, excluded from the normal Tab sequence (the listbox pattern). Reworked as a real `<button>`, appended as a sibling of the results list rather than a row inside it: always visible without scrolling, and reached with a single Tab from the search field (option rows are `tabindex="-1"` and skipped).
- **Activated by a right click.** selectWoo's own option-row activation binds on plain `mouseup` with no mouse-button check at all, so a right click fired the same activation a left click did. Moving the affordance out of the results list and into a real `<button>` removes it from that binding entirely — a `<button>`'s native `click` event only ever fires for the primary mouse button, and Enter/Space work for free.
- **Keyboard input gets swallowed after tabbing away from an open dropdown.** `waitToFocus` (the poll that lands focus in the dropdown's search field after it opens, needed because the picker's own focus-on-open doesn't land reliably on every host theme) kept nudging focus back into the search field for its whole window — up to ~4.8s, or ~12.8s from the results-render-triggered variant — regardless of what happened after it was scheduled. A buyer who Tabbed to a completely different field got yanked back into the dropdown until the poll's hit count ran out or they hit Esc. Scoped the poll to stop the moment focus moves outside the company-search widget.
- **Placeholder styling.** "Search for company" rendered upper-cased (a host theme's blanket `text-transform`) and a few pixels off vertical centre (a host theme's custom selection-box height with no matching line-height). Neutralised the transform and switched to flex centring, scoped to this one field so it doesn't touch any other select2 instance a theme already renders correctly.

Not fixed in this PR: the report that the company-search box itself sits last in the page's DOM/tab order. I could not find a cause for that in this repo's field-priority code (`WC_Twoinc_Checkout::update_company_fields` already assigns it the same priority as WooCommerce's own `billing_company` field), so it may be theme- or overlay-specific rather than a bug here — flagging rather than guessing at a fix.

## Test plan

- [x] `npm run test:js` — 155 passing, including a new focus-trap regression suite and a substantially reworked manual-entry suite (real widget via jsdom + selectWoo, no mocks)
- [x] `php tests/unit/run.php` — all passing
- [x] Mutation-verified: reverted the button rework to the old `<li>` and to an in-list append, confirmed the relevant tests fail; reverted the focus-trap fix, confirmed the new focus tests fail
- [x] `prettier --check` clean on all touched JS/CSS/test files

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -168,35 +168,34 @@
 }
 
 /*
- * Manual-entry row inside the company-search results list (TWO-25288).
+ * Manual-entry button, a sibling of the results list rather than a row
+ * inside it (#30.x.1-3, superseding the TWO-25288 pseudo-option).
  *
- * It is an option row now rather than a floating div beside the list, so the
- * hand-rolled margin is gone: the row inherits the picker's own option padding
- * and highlight, which is what keeps it in step with the company rows above it
- * when arrowed onto. Kept as a flat single-id selector so brand overlays that
- * recolour it keep matching without raising their own specificity.
+ * A real <button> now, so its native chrome has to be reset to read as the
+ * link it visually is — the same treatment #search_company_btn already gets
+ * below. `display: block` + full width keeps it filling the dropdown the way
+ * the old option row did; the padding roughly matches selectWoo's own option
+ * row padding so it doesn't look like a foreign element bolted underneath.
+ * Kept as a flat single-id selector so brand overlays that recolour it keep
+ * matching without raising their own specificity.
  */
 #company_not_in_btn {
+  display: block;
+  width: 100%;
+  text-align: left;
   cursor: pointer;
   color: #3043d1;
+  background: none;
+  border: 0;
+  border-top: 1px solid #eeeeee;
+  font: inherit;
+  padding: 6px 12px;
 }
 
-/*
- * Highlighted state (TWO-25288). NOT cosmetic — a contrast fix.
- *
- * The picker paints its highlighted row white-on-blue, but its own rule is a
- * class selector (specificity 0,3,1) and the id rule above is 1,0,0, so the
- * brand blue above wins on the text while the blue background still applies:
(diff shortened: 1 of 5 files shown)


dgjlindsay:
## Self-review (one adversarial round)

- Focus-scoping fix (`focusStillWithinCompanySearch`): verified against selectWoo's real markup that the explicit search-input check is necessary (the search field is a sibling of `.select2-results`, not a descendant of the results list), and that arrow-navigated option rows and the collapsed combobox trigger are still correctly treated as "inside the widget". No case found where it either steals focus from a field the buyer tabbed to, or wrongly aborts the initial focus-landing on open.
- No dangling references to the removed `<li>`-based mechanism (`manualEntrySentinelId`, `select2:selecting` interception, `manualEntryObserver`, old `--selectable`/`--highlighted` CSS) anywhere in js/css/tests/php.
- No functional collision with open PR #415 (touches a different function/CSS region — the company-summary display, not the manual-entry button or `waitToFocus`). Textual proximity in `twoinc.css` makes a merge conflict likely when #415 lands later; not a defect in this PR.
- New button CSS renders as a full-width dropdown row (not a foreign element).
- Minor, non-blocking: the "right click doesn't activate it" test fires a synthetic event on the button directly rather than through selectWoo's real `mouseup` delegation, so it doesn't itself exercise the old library-level bug path — the button-tag-type and outside-the-list structural tests (both mutation-verified) are what actually catch a regression on this bug class.

Round found nothing that needs a code change. Merging.

[Claude Sonnet 5]

PR #417, opened 2026-07-31
Cache-bust enqueued assets by file mtime

dgjlindsay:
## Summary
- Static asset URLs (`assets/js/twoinc.js`, `assets/css/twoinc.css`, `assets/js/admin.js`, `assets/css/admin.css`) were versioned with the plugin's static version string, which only changes on a deliberate version bump. Between bumps, a deploy that changes an asset's contents did not change its `?ver=` query arg, so CDN/browser caches kept serving the stale file after deploy.
- Adds `twoinc_get_asset_version()`, which derives each asset's version from its own `filemtime()`, falling back to the plugin version string if the file can't be stat'd. Every `wp_enqueue_script()`/`wp_enqueue_style()` call in the plugin now uses it.

## Test plan
- [x] `php tests/unit/run.php` — added `testAssetVersionTracksFileMtimeNotPluginVersion` (two real asset files resolve independent, mtime-derived versions) and `testAssetVersionFallsBackToPluginVersionWhenFileMissing` (missing file falls back cleanly, no warning). Full suite passes.
- [x] `php -l` on all four changed files.

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -2406,8 +2406,8 @@ if (!class_exists('WC_Twoinc')) {
                 wp_enqueue_media();
             }
 
-            wp_enqueue_script('twoinc.admin', WC_TWOINC_PLUGIN_URL . '/assets/js/admin.js', ['jquery']);
-            wp_enqueue_style('twoinc.admin', WC_TWOINC_PLUGIN_URL . '/assets/css/admin.css');
+            wp_enqueue_script('twoinc.admin', WC_TWOINC_PLUGIN_URL . '/assets/js/admin.js', ['jquery'], twoinc_get_asset_version('assets/js/admin.js'));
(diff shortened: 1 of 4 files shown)


dgjlindsay:
**Self-review (adversarial pass, one round — nothing new found):**

- Coverage: every `wp_enqueue_script()`/`wp_enqueue_style()` call for a plugin-owned asset now versions via `twoinc_get_asset_version()` (main CSS/JS + admin CSS/JS). The one remaining unversioned enqueue (`jquery-ui-datepicker`) is a WordPress core script, out of scope.
- Missing-file safety: falls back to the plugin version string (no PHP warning, no empty `?ver=`) — covered by `testAssetVersionFallsBackToPluginVersionWhenFileMissing`.
- Performance: a `filemtime()`/`file_exists()` stat pair per asset, at most 2 per storefront page load — negligible for this plugin's asset count; no transient-caching of the cache-buster warranted.
- CI: all checks green except `e2e-tests`, which fails identically on `staging` HEAD right now (GCP `secretmanager.versions.access` PERMISSION_DENIED for `gha-e2e-tests@two-beta`) — pre-existing infra issue, unrelated to this diff, not fixable from this PR.

by Claude

PR #418, opened 2026-07-31
fix(company-search): uppercase/alignment round 2 + Tab-to-button shortcut

dgjlindsay:
## Summary

Round 2 of the company-search widget fixes from PR #416, reported live off staging.

- **Uppercase (#30.x.5.1/.2)**: `text-transform: none` on the placeholder/selected-value text had no `!important`, so a host theme's own `!important` rule still won. Also added `text-transform: none !important` to `#company_not_in_btn` — as a real `<button>` (since #416) it now matches a theme's generic `button { text-transform: uppercase }` styling, which an `<li>` never did.
- **Vertical alignment (#30.x.5.3)**: round 1 flex-centred only the parent `.select2-selection--single` box; the rendered text's own box wasn't a flex container, so a theme-supplied height with a mismatched line-height still left text off-centre. `.select2-selection__rendered` is now a flex container in its own right with `height: 100%`.
- **3px right padding (#30.x.5.4)**: added to the placeholder text.
- **Tab-to-button (#30.x.6)**: while the dropdown is open and the search field has focus, Tab now moves focus straight to the "not on the list" button instead of tabbing out of the widget entirely. Shift+Tab is untouched (normal reverse-tab).

Root cause of the Tab fix taking two attempts to land locally: selectWoo's own `$(document).on('keydown', ...)` treats a bare Tab exactly like Enter while the dropdown is open, and unconditionally refocuses the search field afterward with no `isDefaultPrevented()` check. `preventDefault()` alone doesn't stop it — `stopPropagation()` does, since that document-level handler is bound above our `document.body` delegation in the bubble chain.

## Test plan

- [x] `npm run test:js` — 159/159 passing, including the existing `company-search-focus-trap.test.js` suite (from #416) confirming the #30.x.4 keyboard-trap fix is untouched.
- [x] New tests in `tests/js/company-search-manual-entry.test.js` under "Tab-to-button shortcut (#30.x.6)": Tab-while-open moves focus to the button, Shift+Tab is left alone, Tab is a no-op below the search threshold (button absent), and an explicit no-trap regression check — dropdown closed, Tab elsewhere on the page stays plain default browser behaviour.
- [ ] Visual check on staging once merged (CSS specificity/theme interaction can't be fully proven from unit tests alone).

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -178,6 +178,15 @@
  * row padding so it doesn't look like a foreign element bolted underneath.
  * Kept as a flat single-id selector so brand overlays that recolour it keep
  * matching without raising their own specificity.
+ *
+ * `text-transform: none` is new (round 2, #30.x.5). It was never needed while
+ * this was an `<li>` — this <button> tag is what makes it a target for a host
+ * theme's generic `button { text-transform: uppercase; }` styling, which
+ * plenty of themes apply for CTA-style buttons. An `<li role="option">` never
+ * matched that selector; a real `<button>` always will. `!important` because
+ * that kind of theme rule is routinely `!important` itself (buttons are a
+ * common place themes fight plugins for the last word), and this element has
+ * no reason to ever want the theme's button casing.
  */
 #company_not_in_btn {
   display: block;
@@ -187,6 +196,7 @@
   color: #3043d1;
   background: none;
   border: 0;
+  text-transform: none !important;
   border-top: 1px solid #eeeeee;
   font: inherit;
   padding: 6px 12px;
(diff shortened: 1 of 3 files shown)


dgjlindsay:
## Adversarial self-review summary (by Claude)

Three rounds of adversarial review (personas: Leia/staff-eng, Han/SRE, Yoda/framework-specialist, Vader/QA) against this PR, plus fixes in response.

**Round 1** (initial diff): no blockers. Nits addressed — `!important` scoping documented, `e.which` vs `e.key` switched to match the vendored selectWoo bundle's own convention, a `height:100%`-on-indefinite-parent CSS caveat and a `.focus()`-on-hidden-element caveat documented as accepted risk (can't verify against arbitrary host themes from here).

**Round 2** (found via Han, reproduced before fixing): BLOCKER — selectWoo's own document-level keydown handler schedules a `focusOnActiveElement()` call 1000ms after every typing keystroke, refocusing the auto-highlighted first result row. A buyer typing and then using the new Tab-to-button shortcut within that ~1s window would see focus land on the button and then get silently yanked back onto a company row. Fixed with a bounded, guarded re-assertion (only fires if selectWoo's timer actually won and the button is still present). New regression test added; also fixed two round-1 tests that had been asserting the wrong thing about `isDefaultPrevented()` (selectWoo already prevents default on any Tab while its dropdown is open, regardless of Shift — pre-existing, not this PR's contract).

**Round 3** (found via Han, reproduced before fixing): BLOCKER — pressing Tab a *second* time, this time from the button, still bubbled to selectWoo's own document handler (its `isOpen()` check is a CSS class, independent of focus) and silently fired `results:select` on the highlighted row before yanking focus back to the search field. Fixed with a second, narrowly-scoped `stopPropagation`-only handler (no `preventDefault`, so native Tab traversal still proceeds) on the button itself. New regression test proves no silent selection.

**Known, deliberately unfixed gap** (surfaced, not introduced, by round 3's fix): selectWoo never clears its own `isOpen()` state on keyboard-only focus-away — before this PR, Tab could never actually escape an open dropdown at all, so nobody could reach "focus moved on, dropdown still open" via keyboard. This PR is the first code path that can reach that state, meaning a buyer who tabs all the way through and away, without ever clicking, can leave selectWoo's global Tab/Enter hijack live for the rest of the page until a stray click closes it. Considered calling `.select2('close')` to close it explicitly; rejected because selectWoo's own `close` handler unconditionally refocuses the collapsed combobox 1ms later, which would reintroduce the exact keyboard trap PR #416 (#30.x.4) was written to fix, one level further out. Documented in code, flagged here as a candidate follow-up ticket rather than patched blind.

Final state: 161/161 JS tests passing, stable across repeated runs; all CI checks green.

PR #419, opened 2026-07-31
fix(company-search): round 3 uppercase, vertical align, Enter/Space activation (#30.x.5, #30.x.6)

dgjlindsay:
## Summary

Round 3 of the company-search widget fixes (TWO-25288, #30.x.5/#30.x.6), reported live off staging after PR #416 and PR #418.

- **Uppercase, still on "Search for company" (#30.x.5.1)**: round 2 added `text-transform: none !important` to `#company_not_in_btn` only, on the correct reasoning that becoming a real `<button>` (since #416) made it a target for a host theme's own `button { text-transform: uppercase }` styling (confirmed live: Astra's own rule list includes the bare `button` element selector, `!important`). That reasoning applies exactly as much to `#search_company_btn` — it has been a real `<button>` since the same PR — but the override was never added there. Same fix, same reason, applied to the button that was missed.
- **Vertical alignment, still off (#30.x.5.3)**: `#search_company_btn` positions `absolute` against `#billing_company_field`, which wraps BOTH the "Company name" `<label>` and the input — only the input has a visible border. With no explicit `top`, the button fell back to its "static position" (roughly after label+input combined in normal flow) rather than the middle of the bordered input box beneath the label. Now appended into `.woocommerce-input-wrapper` (WooCommerce core's own wrapper around just the `<input>`, no label inside it) and centred (`top: 50%; transform: translateY(-50%)`) against that instead — lines up with the visible field regardless of label height or how many lines it wraps to.
- **Enter/Space activation on "not on the list" button doesn't work (#30.x.6)**: round 2 shipped Tab-reachability, but the button's own `keydown` handler only ever intercepted Tab (`e.which === 9`). selectWoo's own document-level keydown handler is gated purely on `isOpen()` — a CSS class, independent of focus — and treats Enter/Space arriving anywhere on the page, including on this button, as "accept the highlighted result": it silently selects whatever company row is currently highlighted and refocuses the search field. That's exactly Doug's report of Enter/Space "routing to the search field". Extended the same `stopPropagation`-only (no `preventDefault`) treatment already used for Tab to Enter and Space, so selectWoo's handler never sees the event but the browser's own native button-activation default action still runs.

## Test plan

- [x] `npm run test:js` — 168/168 passing. New/updated coverage:
  - `#search_company_btn` and `#company_not_in_btn` both declare `text-transform: none !important` (textual assertion against the shipped stylesheet — jsdom does not reliably resolve `!important` + specificity across two separate `<style>` sheets, confirmed directly by a false failure against already-working round-2 code, so this is a textual proof rather than a rendered one).
  - `#search_company_btn` is appended into `.woocommerce-input-wrapper`, not `#billing_company_field` directly, and that wrapper is the positioned containing block the button centres against.
  - Enter and Space, with the button focused, are not hijacked into silently selecting the highlighted result row or refocusing the search field.
- [x] `php tests/unit/run.php` — all passing (unrelated PHP-8.5 `ReflectionMethod::setAccessible` deprecation notices only, pre-existing).
- [x] `prettier --check` clean on all touched files.

## Root cause note

None of these three bugs were new regressions from round 2's changes — they were gaps round 2 didn't cover: the uppercase override was applied to one button but not its sibling, the centring fix targeted the wrong containing element for this specific button, and the Tab-scoped keydown guard was never widened to the other two activation keys. No competing overrides were added on top of round 1/2's fixes; each fix here replaces or extends the existing mechanism directly.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -212,17 +212,46 @@
  * The link back out of manual entry. A <button> now, for keyboard
  * reachability, so the browser's button chrome has to be reset for it to look
  * like the link it did before.
+ *
+ * `text-transform: none !important` is new (round 3, #30.x.5.1). Round 2 only
+ * added this to #company_not_in_btn, on the reasoning that becoming a real
+ * `<button>` (since #416) made it a target for a host theme's own
+ * `button { text-transform: uppercase }` styling. That reasoning applies
+ * exactly as much to THIS button — it has been a real `<button>` since the
+ * same PR — but the override was never added here, so a theme's `!important`
+ * button rule kept winning on this element specifically while the other one
+ * was already fixed. Same fix, same reason, just missed the first time.
+ *
+ * Vertical centring against `.woocommerce-input-wrapper`, not against this
+ * button's own containing `#billing_company_field` (round 3, #30.x.5.3).
+ * `#billing_company_field` wraps BOTH the "Company name" <label> and the
+ * input, but only the input carries a visible border — the label sits above
+ * it with no box of its own. This button used to be positioned with no
+ * explicit `top` at all, so its vertical position fell back to its "static
+ * position" (roughly where it would render in normal flow, i.e. after the
+ * label+input together) rather than to the middle of the bordered input box
+ * beneath the label. `.woocommerce-input-wrapper` is WooCommerce core's own
+ * wrapper around just the <input> — no label inside it — so this button is
+ * now appended there (see `getSearchCompanyBtnNode` in twoinc.js) and
+ * centred with `top: 50%; transform: translateY(-50%)` against THAT box,
+ * which lines it up with the visible field regardless of label height or how
+ * many lines it wraps to on a narrow viewport.
  */
+#billing_company_field .woocommerce-input-wrapper {
+  position: relative;
+}
 #search_company_btn {
   text-align: right;
   cursor: pointer;
-  padding: 5px 0 0;
   color: #3043d1;
   position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
   right: 4px;
   background: none;
   border: 0;
   font: inherit;
+  text-transform: none !important;
 }
 
 /* Payment terms chip selector (TWO-24751) */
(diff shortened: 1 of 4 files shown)


dgjlindsay:

**Round 1** — 4 personas (Leia/maintainability, Han/SRE, Vader/QA, Yoda/framework), run against the initial diff.

Convergent findings (fixed in `5a46a41`):
- **BUG** (Han + Vader): `getSearchCompanyBtnNode()`'s fallback, when `.woocommerce-input-wrapper` was missing, still centred the button against `#billing_company_field` — which already carries `position: relative` from before this round — silently reproducing the exact off-centre bug this PR fixes, untested. Fixed by self-healing (wrapping `#billing_company` in an equivalent span) instead of degrading.
- **NIT** (Han + Vader + Yoda, 3-way): a comment/test description claimed selectWoo's document-level keydown handler treats Space "exactly like Tab" (`results:select`). Checked against the vendored bundle: only Enter/Tab hit that branch; Space only hits the generic focus-steal fallthrough. The fix itself was already correct — only the stated mechanism was wrong. Comment corrected.

Individual findings (fixed in `d3cc213`):
- **Test quality** (Leia): a specificity test only checked the selectors were flat IDs via regex, never actually comparing against the real theme selector list — dropped in favour of the two text-transform tests that already cover the real behaviour.
- **NIT** (Leia): a rationale essay about `.woocommerce-input-wrapper` centring had drifted into three near-identical copies (CSS/JS/test comments). Consolidated to the CSS comment (canonical, next to the rule); JS/test comments shortened to reference it.

**Round 2** — confirming pass verifying the round-1 fixes landed correctly and introduced nothing new. Verdict: **clean**. `npm run test:js` 168/168 passing throughout.

**CI note**: `e2e-tests` is failing on this PR, but it also fails on `staging` HEAD's own history independent of this branch (verified: a `staging`-branch run at commit `740b57f3` failed the same job for unrelated infra reasons — GCP secret-access permission denial in one run, a 404 on an admin-login endpoint in another). Neither failure mode touches anything in this diff (checkout CSS/JS only). No branch protection requires this check on `staging`. Proceeding to merge per the self-review lane.

— by Claude

PR #420, opened 2026-07-31
fix(company-search): focus visibility and Enter/Space activation on the "Search for company" button (#30.x.7)

dgjlindsay:
## Summary

Round 4 of the company-search widget fixes (TWO-25288, #30.x.7), reported live off staging after round 3 (PR #419) merged.

Doug's report: tabbing out of the manual-entry "Company name" field moves focus away from the input, but it's invisible where it lands — assumed to be `#search_company_btn` ("Search for company"). If so: (1) it doesn't show it's focused, and (2) Enter/Space don't activate it. Verified both.

- **(1) No visible focus (#30.x.7)**: `#search_company_btn` carried no host-supplied focus styling and no explicit one of our own — the same class of problem as the uppercase fix (#30.x.5.1): a real `<button>` is a target for a host theme's own button-chrome reset (Astra strips the native focus ring the same way it uppercases button text), which the `<div>` this used to be never was. Added an explicit `!important` outline on `#search_company_btn:focus`, same escalation as the `text-transform` override.
- **(2) Enter/Space don't activate it**: unlike `#company_not_in_btn`'s equivalent fix (round 3, #30.x.6), the interference here cannot be selectWoo's document-level handler — selectWoo's widget is destroyed before this button is ever shown (see `enterManualCompanyEntry`). The native "Enter/Space activates a focused button" default action is presumably being intercepted by something external to this plugin (most likely WooCommerce core's own checkout-submit guard, which treats any `:input` — buttons included — the same way it treats text fields). Rather than chase an unknown, theme-variable interferer, `getSearchCompanyBtnNode` now binds a keydown handler directly on the button element itself — guaranteed by the DOM's own target-then-bubble dispatch order to run before any ancestor-bound handler regardless of where it lives or when it was registered — and drives the switch back to search directly instead of depending on native activation.

## Test plan

- [x] `npm run test:js` — 172/172 passing. New coverage: explicit `!important` outline on `#search_company_btn:focus` (textual assertion — jsdom does not reliably resolve `!important` + specificity across separate stylesheets, per the round-3 precedent in this suite), Enter activates the button, Space activates the button, other keys don't.
- [x] `php tests/unit/run.php` — all passing.
- [x] `prettier --check` clean on all touched files.

## Root cause note

Neither bug is a regression from round 3 — round 3 fixed Tab/Enter/Space for the OTHER button (`#company_not_in_btn`, inside the open search dropdown); this button (`#search_company_btn`, inside manual entry) was never touched by that fix and has different surrounding mechanics (no selectWoo widget alive at all), so it needed its own investigation and its own fix rather than a copy-paste of round 3's.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -254,6 +254,22 @@
   text-transform: none !important;
 }
 
+/*
+ * Explicit focus indicator (#30.x.7, round 4). Reported live: tabbing out of
+ * the manual-entry name field lands focus here with no visible sign of it.
+ * The button carries no host-supplied focus styling of its own — same class
+ * of problem as the uppercase fix above (#30.x.5.1): a real <button> is a
+ * target for a host theme's own button chrome (Astra strips the native
+ * focus ring from buttons as part of the same reset that uppercases them),
+ * where the <div> this used to be never was. `!important` for the same
+ * reason the text-transform override needs it — a theme's own button-focus
+ * reset is routinely `!important` itself.
+ */
+#search_company_btn:focus {
+  outline: 2px solid #3043d1 !important;
+  outline-offset: 2px;
+}
+
 /* Payment terms chip selector (TWO-24751) */
 /*
  * Heading above the chips, shown only when more than one term is offered
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -259,11 +259,15 @@
  * the manual-entry name field lands focus here with no visible sign of it.
  * The button carries no host-supplied focus styling of its own — same class
  * of problem as the uppercase fix above (#30.x.5.1): a real <button> is a
- * target for a host theme's own button chrome (Astra strips the native
- * focus ring from buttons as part of the same reset that uppercases them),
- * where the <div> this used to be never was. `!important` for the same
- * reason the text-transform override needs it — a theme's own button-focus
- * reset is routinely `!important` itself.
+ * target for a host theme's own button chrome, where the <div> this used to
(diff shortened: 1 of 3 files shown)


dgjlindsay:

**Round 1** — 3 personas (Han/SRE, Vader/QA, Yoda/framework). No BLOCKER/BUG. Convergent nit (Han + Yoda): a comment overstated confidence in an unverified root-cause theory (WooCommerce core's checkout Enter-guard) — Yoda noted such guards are typically scoped to input selectors and may not even match a `<button>`. Individual nits from Vader/Yoda on test-intent clarity and a "structurally guaranteed" claim that only holds for bubble-phase listeners. All folded into `14bd965`: hedged both unconfirmed-theory comments explicitly, softened the bubble/capture-phase claim, clarified two test descriptions (one is a selectivity guard, not standalone proof; `.focus()` calls mirror real usage but aren't load-bearing since the handler doesn't gate on focus), and added a forward-reference note in round 3's sibling handler.

**Round 2** — confirming pass verifying the round-1 wording fixes landed correctly and that the follow-up commit touched only comments/test descriptions, not production logic (confirmed via diff). Verdict: **clean**. `npm run test:js` 172/172 passing throughout.

**CI note**: `e2e-tests` failed again on this PR with the same "admin login failed: 404" error already confirmed (in PR #419) to be a pre-existing failure on `staging` HEAD's own history, unrelated to this diff (checkout CSS/JS only). No branch protection requires this check. Proceeding to merge per the self-review lane.

— by Claude

PR #421, opened 2026-07-31
fix(company-search): restyle focus indicator to a dotted border (#30.x.7)

dgjlindsay:
## Summary

Round 5 follow-up on TWO-25288 (#30.x.7), post-merge feedback on PR #420's fix.

Round 4's focus indicator worked (visible, functional) but the styling was wrong: an outline ~4px wider than the button's own padding, inconsistent with this checkout's other focus states (a plain dotted rectangle with square corners).

- Replaced `#search_company_btn:focus`'s `outline` with `border: 1px dotted grey !important` — a flush border rather than an offset outline, matching the checkout's existing focus convention.
- Added explicit `padding: 0 2px` to the base `#search_company_btn` rule (previously relying on the browser's own default `<button>` padding), so the border sits close against the text rather than around an oversized default box.

## Test plan

- [x] `npm run test:js` — 173/173 passing. Updated the round-4 focus test to assert the new border declaration; added a new test for the explicit padding.
- [x] `php tests/unit/run.php` — all passing.
- [x] `prettier --check` clean.

## Root cause note

Not a regression — round 4's fix worked functionally (visible + keyboard-activatable), it was purely a visual mismatch with the checkout's design language, caught in live review after merge.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -250,28 +250,37 @@
   right: 4px;
   background: none;
   border: 0;
+  padding: 0 2px;
   font: inherit;
   text-transform: none !important;
 }
 
 /*
- * Explicit focus indicator (#30.x.7, round 4). Reported live: tabbing out of
- * the manual-entry name field lands focus here with no visible sign of it.
- * The button carries no host-supplied focus styling of its own — same class
- * of problem as the uppercase fix above (#30.x.5.1): a real <button> is a
- * target for a host theme's own button chrome, where the <div> this used to
- * be never was. Assumed, not confirmed live in devtools the way the
- * uppercase fix's Astra rule was (#30.x.5.1's comment quotes the actual
- * selector list found) — plausible that Astra's same button reset also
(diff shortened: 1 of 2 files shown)


dgjlindsay:
## Adversarial self-review — 1 round (lighter review, per Doug)

Single-reviewer QA pass (small, purely-visual CSS diff). No blockers. Two optional nits folded in via a follow-up commit:
- Drawing the focus border only on `:focus` grew the button's box by 2px on focus/blur, nudging its unfixed-width, right-pinned position ~1px sideways each way. Reserved the same border width/style (transparent) in the base rule so `:focus` only ever changes colour, never box size.
- `grey` was the only named CSS color in the stylesheet; switched to `#808080` to match the file's hex convention.

`npm run test:js` 174/174 passing throughout, `prettier --check` clean, `php tests/unit/run.php` passing.

Merging on green per Doug's call — no confirming round, no 10-minute wait for this one.

— by Claude

PR #422, opened 2026-07-31
fix(checkout): stop billing field reorder, move optionals to the bottom

dgjlindsay:
## Summary
- Stop `positionFields()` pulling `billing_phone`/`billing_email` into the pre-billing "representative" wrapper — this ~100ms move was the cause of the ~1s visible field reorder (fields jumped from server-rendered order into a different order once JS ran).
- That wrapper's own visual cue (an `<h3>` heading) was commented out in 2021 with no CSS replacement, so the grouping has no remaining display purpose today. First/last name still move there (a no-op position-wise, since they're already first).
- Move the optional fields (invoice email, PO number, project, department) off company's own priority onto a fixed baseline below phone/email, so they always land at the bottom regardless of company field state.

Final order: first/last name -> country -> company -> street address x2 -> postcode/city -> phone -> email -> optional fields.

## Test plan
- [x] `make test-unit` — all pass, including new `testFieldPrioritiesMatchDesiredCheckoutOrder`
- [x] `npm run test:js` — 193 passed
- [x] Live-verified via Chrome against a local dev checkout (company search + all optional fields enabled): billing field order identical between initial paint and t+2s — no reorder

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -888,7 +888,15 @@ let twoincDomHelper = {
   },
 
   /**
-   * Move the fields to their original or Twoinc template location
+   * Move the fields to their original or Twoinc template location.
+   *
+   * Phone and email used to be pulled up here too (into the pre-billing
+   * "representative" wrapper, alongside first/last name), so a buyer would
+   * see one field order on first paint and a different one ~1s later once
+   * this fired. That grouping's own visual cue (an h3 heading) was commented
+   * out back in 2021 and never replaced with CSS, so nothing distinguishes
+   * the wrapper today — it was pure reorder with no remaining display
+   * purpose. Phone/email now stay in their native WC position (#33).
    */
   positionFields: function () {
     setTimeout(function () {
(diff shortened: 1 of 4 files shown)


dgjlindsay:
## Adversarial self-review (round 1) — by Claude

**Han (regression risk, representative-fields JS/view change):** no BLOCKER/BUG. Verified guest/non-Two path, company-search-off, order-pay page (`views/woocommerce_order_pay.php`, fully independent markup), no other JS/CSS/test refs to the removed `twoinc-ph-target`/`twoinc-em-target` targets, no CSS sized for the old 4-field wrapper.

**Vader (edge cases, priority-baseline change):** found one latent (not currently exploitable) issue — the new fixed optional-field baseline (200) decoupled `invoice_email`/PO/project/department from company's own priority; a future brand overlay pushing `billing_company`'s priority to 196+ could have inverted company/company_id above the optionals. No such overlay exists today, but the diff introduced the risk. **Fixed in follow-up commit 547c004**: clamped company's priority read at 190 in both `move_country_field()` and `update_company_fields()`, with a regression test (`testCompanyPriorityClampPreventsInversionAboveOptionals`).

Round 2 not run — Han's findings were all clean, Vader's one finding is fixed and test-covered. No new findings expected from a repeat pass on an unchanged surface.

dgjlindsay:
## Adversarial self-review (round 2, confirmation) — by Claude

Re-ran Vader against commit 547c004 (the clamp fix for round 1's finding) with a confirm-only mandate.

**Verdict: CONFIRMED CLEAN.**
- `move_country_field()` and `update_company_fields()` both compute `min($fields['billing']['billing_company']['priority'] ?? 30, 190)` before deriving country/company/company_id. Verified against inputs 1000, 200, 199, 191, 190, 30 (default), and negative — all resolve strictly below the fixed optional-field baseline (200).
- New test `testCompanyPriorityClampPreventsInversionAboveOptionals` forces `billing_company.priority = 1000` and asserts against the real production methods — not tautological, fails without the fix.
- No brand overlay in `brands/` sets a `billing_company` priority in the 190-199 range — clamp breaks no existing legitimate use.
- Fresh pass over the full diff surfaced nothing new in the JS/view change.

Two clean confirming rounds. No outstanding findings.

PR #423, opened 2026-07-31
fix(TWO-33): sync WooCommerce's locale defaults with the country/company priority clamp

dgjlindsay:
## Summary

`#33` shipped a server-side fix (PR #422 / commit 547c004) so country renders above company on checkout. It works for the raw HTML — but WooCommerce core's own `address-i18n.js` independently re-derives `#billing_country_field`'s priority from `WC_Countries::get_country_locale()`'s `default` entry on **every** checkout load (not only on country change), then physically re-sorts the billing form's DOM. That locale array is untouched by this plugin's `woocommerce_checkout_fields` filters, so it still carries WooCommerce's hardcoded defaults (company priority 30, country priority 40) and silently re-inverts the order client-side a few hundred ms after load — undoing the #33 fix in the browser only.

Confirmed directly against the real `woocom-dev.staging.two.inc` deployment: server-rendered HTML had country above company; the live DOM did not, because this second, independent mechanism (WooCommerce's own bundled JS) reset it back. Never reproducible against a local fixture, since it depends on that bundled JS reading server-localized JSON rather than anything this plugin executes.

## Changes

- Adds `sync_locale_country_priority()`, hooked on `woocommerce_get_country_locale_default`, applying the same company/country clamp used by `move_country_field()`/`update_company_fields()`.
- Extracts the shared clamp into `clamp_company_priority()` so all three call sites can't drift apart.
- Strengthens the field-order regression test to assert country's priority is strictly below **both** `billing_company` and `billing_company_display`.
- Adds a new unit test covering `sync_locale_country_priority()` directly (including the 190 ceiling).

## Test plan

- [x] `make test-unit` (PHP 8.2, same as CI) — all tests pass, including the two new/updated ones
- [x] `npm run test:js` — 193 tests pass (no JS changes; included for completeness)
- [x] Live-verified root cause and (pending merge + git-sync) will re-verify the fix against the real staging URL

### DIFF
--- a/class/WC_Twoinc_Checkout.php
+++ b/class/WC_Twoinc_Checkout.php
@@ -27,6 +27,20 @@ if (!class_exists('WC_Twoinc_Checkout')) {
             add_filter('woocommerce_checkout_fields', [$this, 'add_tracking_fields'], 21);
             add_filter('woocommerce_checkout_fields', [$this, 'update_company_fields'], 23);
 
+            // WooCommerce's own address-i18n.js re-derives #billing_country's
+            // client-side priority from THIS locale default array on every
+            // checkout load (fired via country_to_state_changing at init, not
+            // only on country change) and then re-sorts the DOM — entirely
+            // independent of the woocommerce_checkout_fields chain above.
+            // Without this mirror, move_country_field()'s server-side fix is
+            // silently undone a few hundred ms after page load: JS resorts
+            // .form-row elements using the hardcoded core default
+            // (country priority 40) against company's 30, putting company
+            // back above country (#33 — live-staging-only regression, never
+            // reproduced against a local fixture because it depends on
+            // WooCommerce's own bundled JS, not this plugin's).
+            add_filter('woocommerce_get_country_locale_default', [$this, 'sync_locale_country_priority']);
+
             // Brand overlays add/modify checkout fields after the base set
             // is complete (priority 25 > the base mutations above)
             add_filter('woocommerce_checkout_fields', [$this, 'apply_brand_checkout_fields'], 25);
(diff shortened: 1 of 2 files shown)


dgjlindsay:
[Claude Sonnet 5] Self-review summary (Leia/Han/Vader personas, round 1):

**Leia (maintainability)** — no blockers. Minor naming nit accepted implicitly (kept `clamp_company_priority` name, judged clear enough given the docblock cross-references).

**Han (SRE/regression-risk)** — verified `woocommerce_get_country_locale_default` is also hooked by WooCommerce Blocks' `CheckoutFields::update_default_locale_with_fields()`, but that callback only fills in *missing* keys (`if (empty($locale[$field_key]))`), so it never overwrites our 'country'/'company' values regardless of registration order — not a conflict. Confirmed no brand overlay or other active plugin on the real shop touches `billing_company`'s priority or hooks this filter, so the two priority mechanisms (checkout-fields vs locale-default) can't currently diverge in practice.

**Vader (edge cases)** — two real findings, both fixed in the follow-up commit (`1f757bd`):
- MEDIUM: `sync_locale_country_priority()` blind-wrote `$fields['country']['priority']` with no `isset` guard, which would auto-vivify a malformed entry if 'country' were ever absent from the array. Added a guard + test.
- HIGH (documented, not code-fixed): the locale-default filter reads WC core's static 'company' default, not `billing_company`'s live/brand-adjusted priority — a latent gap if a future brand overlay ever customizes `billing_company`'s priority without also touching locale defaults. No current brand does this; documented in the method's docblock as a known limitation for the next person who adds one.

Round 2 not run — remaining findings were either non-issues (idempotency, PHP type-juggling on `min()`) or the documented latent gap above, which is out of scope for this fix (no active bug today). CI green (phpstan rerun after a transient `libsodium.so.23` runner-infra flake; e2e-tests red on the same admin-login 429 rate-limit flake seen on prior PRs today — pre-existing infra, unrelated to this diff).

PR #424, opened 2026-08-01
fix: company-number label and search-again link overlapping the company field

dgjlindsay:
## Summary

Two live layout bugs Doug found testing woocom-dev.staging.two.inc/checkout/ with a real cart and real company search (TWO-25324):

- **Company number invisible after picking a search result.** `.twoinc-company-summary-id` shared the same line as the company name (`margin-left: 8px`, `white-space: nowrap`), so a long company name pushed the number toward — and on a narrow viewport past — the right edge of the summary box. Fix: the number now renders as its own block row, right-aligned to the input's right edge, decoupled from the name entirely.
- **"Search for company" link rendering on top of the company-name field.** `#search_company_btn` was `position: absolute; top: 50%; transform: translateY(-50%)` against `.woocommerce-input-wrapper` — centred vertically over the input, i.e. visually overlapping it. Fix: normal block flow as the wrapper's last child, right-aligned, no absolute positioning — matching the layout pattern `magento-plugin`'s Luma `.search_for_company` and `magento-hyva-extension` already use.

## Test plan

- [x] `npm run test:js` — 195/195 passing, including 3 new/updated assertions
- [x] Mutation-verified: reverted the CSS to the pre-fix state and confirmed all 3 new/updated assertions fail (not vacuous) — restored, all green again
- [x] Live-verified pre-fix on woocom-dev.staging.two.inc/checkout/: confirmed via devtools that a picked company's summary renders `<span class="twoinc-company-summary-name">` + `<span class="twoinc-company-summary-id">` inline, matching the described bug
- [ ] Live-verify post-merge on staging

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -302,11 +302,27 @@
  * renders inside/over the field" bug. With the absolute positioning removed
  * the button falls into NORMAL block flow as the wrapper's last child — i.e.
  * immediately below the input, inside the same box the input's own border
- * defines, so `text-align: end` lines its text up with the input's right
- * edge without needing any explicit positioning at all. Matches the pattern
- * Magento/Luma (`magento-plugin` `.search_for_company`) and Hyvä already use:
- * a plain block appended after the field, right-aligned, no absolute
- * positioning.
+ * defines. Matches the pattern Magento/Luma (`magento-plugin`
+ * `.search_for_company`) and Hyvä already use: a plain block appended after
+ * the field, right-aligned, no absolute positioning.
+ *
+ * `width: 100%` is still required for `text-align: end` to mean anything
+ * (round 2 review — Vader, correcting round 1): a `<button>` is a form
+ * control, and form controls use intrinsic (shrink-to-fit) sizing at
+ * `width: auto` REGARDLESS of `display: block` — unlike an ordinary
+ * element, which fills its containing block at `width: auto` once
+ * blockified. Round 1 deleted `width: 100%` believing `display: block`
+ * alone was sufficient (the same belief `#company_not_in_btn`'s own
+ * comment above already contradicts: "`display: block` + full width keeps
+ * it filling the dropdown"). Without it, this button shrink-wraps its own
+ * label and sits at the input's LEFT edge — `text-align: end` becomes a
+ * no-op on a box already exactly the text's width, silently reintroducing
+ * a left-aligned link where the design calls for right-aligned. `box-
+ * sizing: border-box` is what round 1's original concern (no global
+ * border-box guaranteed in this stylesheet — verified, zero hits) actually
+ * needed: it keeps `width: 100%` + this rule's own padding/border from
+ * overflowing the containing block by 6px, without giving up the width
+ * declaration itself.
  *
  * The wrapper itself gets an explicit `display: block` (round 1 review —
  * Han/Vader, convergent), replacing the `position: relative` this rule used
@@ -328,6 +344,8 @@
 }
 #search_company_btn {
   display: block;
+  box-sizing: border-box;
+  width: 100%;
   margin-top: 4px;
   text-align: end;
   cursor: pointer;
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -377,9 +377,9 @@
  * deliberately, rather than `border: 0` — a border occupies box space
  * (unlike outline, which doesn't), so drawing it only on `:focus` would grow
  * the button by 2px the moment focus lands and shrink it back on blur,
- * nudging its (unfixed-width, right-pinned) box and its text sideways by
- * ~1px each way. Reserving the same width up front, invisible until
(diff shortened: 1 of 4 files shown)


dgjlindsay:
**Adversarial review — round 1** (Han + Vader personas, two-review-adversarial skill)

Convergent findings, both fixed in 23b47a4:
- `.woocommerce-input-wrapper` (a `<span>`) needs explicit `display: block` — a host theme setting it to `flex` would silently re-create the overlap this PR removes.
- `width: 100%` on `#search_company_btn` overflows its box under content-box sizing; removed (a `display: block` element already fills its containing block at `width: auto`).
- Summary padding now matches WooCommerce core's `.form-row { padding: 3px }` so the right-aligned number lines up with the input's actual edge.

Single-reviewer findings, also fixed:
- Han: pay-for-order page renders the summary as a full-page-width flex row, so right-aligning the number there detached it from the (narrower, centred) field above — same bug class, different surface. Overridden back to the leading edge on `.custom-checkout`.
- Vader: long unbroken company names/org numbers could still overflow the row. Added `overflow-wrap: anywhere`, dropped the now-counterproductive `white-space: nowrap` on the id.
- Vader: `text-align: right` → `text-align: end` (logical, no behaviour change today).
- Vader: `margin-top: 4px` was mutation-passed unasserted — added coverage.

Declined (edge case, pre-existing, out of this PR's scope, not a regression): Vader's note that a number-only summary (name blank, number present) renders as a lone right-aligned value with nothing above it — that visibility gate (`Boolean(name || number)`) predates this PR and isn't part of the reported bug.

All fixes mutation-verified (reverted, confirmed the new/updated assertions fail, restored). Full suite: 199/199 passing.

— Claude

dgjlindsay:
**Adversarial review — round 2** (Han + Vader)

Han: verified all 5 round-1 fixes land as claimed (read current source, confirmed the `.custom-checkout` override wins the cascade empirically via jsdom, ran the suite himself — 199/199). One NIT: the `.custom-checkout` test was regex-only (proves the rule exists, not that it wins). Fixed with a computed-style assertion.

Vader: caught a real regression **introduced by round 1's own fix** — round 1 deleted `width: 100%` on `#search_company_btn` believing `display: block` alone fills the containing block at `width: auto`. True for an ordinary element, **false for a `<button>`** — form controls use intrinsic shrink-to-fit sizing regardless of `display`. Without the width, the button hugs its label and sits at the input's left edge, making `text-align: end` a no-op — silently reintroducing a left-aligned link. Fixed in 5c07b34: restored `width: 100%` paired with `box-sizing: border-box` (which is what round 1's original overflow concern actually needed).

Also closed two mutation-survivable test gaps: the `.custom-checkout` cascade-winner assertion above, and computed-style coverage for `.twoinc-company-summary`'s form-row padding (a naive stylesheet regex would silently pass regardless, since the rule's own CSS comment contains a `}` that truncates a `[^}]*` capture early).

All fixes mutation-verified. Full suite: 202/202 passing. Running round 3 since round 2 found real issues (not a clean round per the no-oscillating-loop rule).

— Claude

dgjlindsay:
**Adversarial review — round 3 (clean)**

Han and Vader both independently ran the suite (202/202 green), re-verified round 1 and round 2's fixes are correct and present in the shipped CSS, checked for any other form-control-auto-fill mistakes elsewhere in this diff (none), and confirmed no other rule re-targets `#search_company_btn` later in the cascade. Vader additionally mutated the two new computed-style tests himself (11 separate mutations, all caught) and enumerated every declaration in every rule this PR touches against test coverage — everything added or modified has coverage; the only uncovered declarations are pre-existing lines this diff carried through untouched.

No bugs in this round. Two comment-accuracy nits (a stale "unfixed-width, right-pinned" description now that `width:100%` ships without absolute positioning, and a test comment that claimed to guard against rule reordering when the override actually wins on specificity regardless of order) — fixed in a4c1c93, no behavior change.

Both reviewers recommended stopping here. Merging.

— Claude

dgjlindsay:
e2e-tests failing on \`Run e2e tests\` — confirmed unrelated to this diff:

\`\`\`
Error: admin login failed: 404 <!DOCTYPE html>...
at triggerFulfilBatch (tests/e2e/two-api.ts:80:11)
at tests/e2e/tests/order-flow.spec.ts:30:3
\`\`\`

This is the same workflow failing independently on \`staging\` itself (run 30617611990, 2026-07-31, headBranch=staging) — a pre-existing infra/environment issue (an external Two API endpoint returning 404 on admin login) unrelated to a front-end CSS/JS layout fix. Reproduced identically across two reruns of this PR's own run, ruling out a one-off flake. Not fixing here — out of scope for this PR and the fix doesn't live in this repo's front-end code.

Merging per the self-review lane: no branch protection, no required status checks, \`reviewDecision\` null, 3 clean adversarial review rounds, 10-minute quiet window elapsed. All checks this PR is actually responsible for (Jest, PHP unit tests 7.4-8.4, phpcs, phpstan, pre-commit, gettext, boot/upgrade smoke) are green.

— Claude

PR #425, opened 2026-08-01
fix: company summary self-heals position after WooCommerce's own field resort

dgjlindsay:
## Summary

Follow-up to PR #424 (TWO-25324). The mandated post-merge live-verify on woocom-dev.staging.two.inc/checkout/ found a real gap: the company-number summary now renders as a correctly right-aligned block (pixel-confirmed against the input's right edge), but the whole summary block sat **above** the company field instead of below it — failing the literal acceptance criterion for #1.1.

Root cause: `WC_Twoinc_Checkout.php` already documents (in the comments above `move_country_field()` and `sync_locale_country_priority()`) that WooCommerce core's own `address-i18n.js` detaches and re-appends every `.form-row` in the billing wrapper by priority, on **every** checkout load — not only on country change. `getCompanySummaryNode()` positioned the summary `<div>` once, on first creation, and never again — but the summary isn't a `.form-row`, so it never takes part in that resort. Once WC moved the real fields past it, it stayed stranded above all of them for the rest of the page's life. The plugin already carries two established fixes for exactly this mechanism (for the country field); this is the same class of bug for the summary.

Fix: re-run the anchor's `insertAfter` on every `getCompanySummaryNode()` call, not just on creation. `insertAfter` on an already-attached node moves it rather than cloning, so any `renderCompanySummary()` call (pick, payment-method switch, country change, re-render — all of which already call it) snaps the summary back into place, at the cost of one no-op DOM move when nothing has drifted.

## Test plan

- [x] `npm run test:js` — 203/203 passing (added 1 new test)
- [x] New test simulates WooCommerce's own field resort (detach + re-append the real fields past the summary) and asserts the next `renderCompanySummary()` call snaps the summary back into place
- [x] Mutation-verified: reverted the fix, confirmed the new test fails, restored
- [ ] Live-verify post-merge on staging

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -1270,11 +1270,23 @@ let twoincDomHelper = {
    * them. The plugin already carries two established fixes for exactly this
    * mechanism (for the country field) — this is the same class of bug for
    * the summary. `insertAfter` on an already-attached node MOVES it rather
-   * than cloning, so re-running it here on every `renderCompanySummary()`
-   * call (which already fires on every pick, payment-method switch, country
-   * change and re-render) snaps the summary back into place after any
-   * external resort, at the cost of one no-op DOM move on calls where
-   * nothing has drifted.
+   * than cloning, so re-checking the anchor here on every
+   * `renderCompanySummary()` call (which already fires on every pick,
+   * payment-method switch, country change and re-render) snaps the summary
+   * back into place after any external resort.
+   *
+   * Guarded on `$node.prev()` (round 1 review — Han): re-running
+   * `insertAfter` UNCONDITIONALLY, on every call, physically detaches and
+   * re-inserts the node even when nothing has drifted — measured with a
+   * MutationObserver, every "healthy" call still fires a childList removal
+   * + addition. That collapses any text selection inside the summary (the
+   * only interaction this read-only org-number display affords is
+   * selecting it to copy), forces a reflow, and would restart any CSS
+   * transition a brand overlay puts on this element (`.custom-checkout
+   * .twoinc-company-summary` in twoinc.css proves overlays do style it).
+   * `.prev()` is element-only (ignores text nodes), so "prev is already the
+   * anchor" reliably implies "already positioned, same parent, nothing to
+   * do" — the move only runs when the anchor actually changed.
    *
    * @returns {Object} jQuery-wrapped summary, or an empty set on a page with
    *   no company fields at all
@@ -1299,7 +1311,8 @@ let twoincDomHelper = {
     }
 
     const $wrapper = $field.closest(".twoinc-inp-container");
-    $node.insertAfter($wrapper.length ? $wrapper : $field);
+    const $anchor = $wrapper.length ? $wrapper : $field;
+    if ($node.prev()[0] !== $anchor[0]) $node.insertAfter($anchor);
     return $node;
   },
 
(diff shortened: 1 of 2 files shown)


dgjlindsay:
**Adversarial review — round 1** (Han + Vader)

Vader: mutation-verified the fix independently (8 mutations against the actual anchor logic — reversion to create-once, insertAfter→insertBefore, priority-swap, etc.), confirmed the new test is the sole killer for the ones that matter. Checked the pay-for-order page's separate `.twoinc-inp-container` anchor path (clean, no interaction), and the field-vanishes-mid-render edge case (no regression, same outcome as pre-fix). No blocking findings, two coverage NITs noted (untested pay-page wrapper path, untested `isNew` branch — both pre-existing/low-risk, not fixed this round).

Han: confirmed the anchor choice (`#company_id_field` after `#billing_company_display_field`) is correct for the right reason — verified the actual WooCommerce field priorities in `WC_Twoinc_Checkout.php` rather than assuming. Found a real issue: repositioning **unconditionally** on every call physically detaches and re-inserts the DOM node even when nothing drifted (measured with a MutationObserver) — collapsing any text selection inside the summary (the only interaction this read-only display affords) and forcing an avoidable reflow. Fixed in 94821db: guard on `$node.prev()` matching the computed anchor, only move when it's actually changed. This also eliminates his related observation about layout shift under an open dropdown.

Declined as out-of-scope hardening (both explicitly optional in the reviewer's own writeup): subscribing to `country_to_state_changed` directly instead of relying on script enqueue order, and anchoring on "last company row in the DOM" instead of a fixed field-priority order.

Mutation-verified the round-1 fix (spied on `insertAfter`, confirmed it's not called on an already-positioned re-render; reverted the guard and confirmed the new test fails). Full suite: 204/204 passing.

— Claude

dgjlindsay:
**Adversarial review — round 2 (clean)**

Han: independently verified the round-1 guard fix is correct in both branches, confirmed the anchor-identity-swap concern isn't reachable (WC's own fragment-refresh replaces the review table/payment box, never the billing wrapper — and if it ever did, the summary would self-heal via recreation), and verified the spy-test's mock cleanup is double-covered (`jest.config.js` already sets `restoreMocks`+`resetMocks`, confirmed on origin/staging pre-dating this PR). One pre-existing, non-blocking coverage gap noted (pay-for-order page's `.twoinc-inp-container` anchor branch has no DOM-level test) — flagged as a follow-up ticket candidate, not fixed here.

Vader: ran 13 independent mutations against the guard logic (revert, `!==`→`===`, `.prev()`→`.next()`, `insertAfter`→`insertBefore`/`appendTo`, anchor-selection changes, etc.) — all real mutations killed by name-appropriate tests, confirmed 2 as equivalent mutants (no behavior difference) and 2 as the same pre-existing gap Han found. One real cosmetic finding: a dead ternary (`isNew ? jQuery() : $node`) that's provably identical to plain `$node` at that line — fixed in b1263ed, no behavior change.

No bugs in this round. Both reviewers recommend merge.

Full suite: 204/204 passing.

— Claude

dgjlindsay:
e2e-tests failing again with the identical signature confirmed on #424 (admin login 404, pre-existing infra breakage unrelated to this diff). All checks this PR is responsible for are green. Merging per the self-review lane (no branch protection, reviewDecision null, 2 review rounds — round 1 found and fixed a real DOM-churn issue, round 2 clean, 10-minute quiet window elapsed).

— Claude

PR #426, opened 2026-08-01
fix: three checkout regressions from #30.x.9 CSS rework (#30.x.13)

dgjlindsay:
## Summary

Doug found 3 new regressions live-testing checkout right after the #30.x.9 company-field CSS fix (#424, #425). All three trace back to the same area (manual company entry).

1. **Search-again button click does nothing.** Live reproduction: the click DOES reach `#search_company_btn` (capture-phase listener confirms target, `defaultPrevented: false`), but never bubbles far enough to reach the `document.body`-delegated click handler — something in the real checkout stops propagation between the button and body. A real Enter keypress on the same button *did* work, because that handler is bound directly on the element. Fix: bind click directly on the button too, alongside its existing Enter/Space handler, so it no longer depends on bubbling reaching body at all.

2. **`#company_id_field` ("Company ID") visible in manual entry mode.** `toggleBusinessFields` decides whether to show this field purely from `window.twoinc.enable_company_search === "no"` — but three different code paths set that to `"no"`: merchant-level search-off (wants the id field), sole-trader mode (wants a synthetic id), and manual entry (name-only, no id, per the org's settled three-mode company-capture model). Added `window.twoinc.manual_company_entry_active` to disambiguate, set/cleared in `enterManualCompanyEntry`/`exitManualCompanyEntry`, and explicitly cleared when entering sole-trader mode.

3. **Page-wide Tab reported broken in manual entry.** Live reproduction (3 consecutive real Tab presses, from an unrelated field, after reaching manual entry via the "not on the list" button) did **not** reproduce the symptom under those repro steps — Tab moved focus normally through the form. However, `enterManualCompanyEntry` calls `select2("destroy")` on a widget that is (by construction) always still *open* at that point, and selectWoo's own close cleanup — which unbinds its document-level Tab/Enter interceptor — only runs on `close()`, not `destroy()`. This is a real, pre-existing gap the code's own comments already flagged. Hardened by calling `close()` before `destroy()`. Flagged for Doug to re-test with his exact repro steps since I could not reproduce the original symptom to confirm this as the fix.

## Test plan

- [x] `npm run test:js` — 206/206 passing (was 205; added 1 net test, see below)
- [x] Mutation-verify: each new/changed test fails against the pre-fix code and passes against the fix (verified via `git stash`)
- [x] Live-verified (real browser clicks/keypresses, not JS `.click()`) via `two-dev-tools-chrome-from-wsl` against staging, before the fix — see PR discussion for raw click-log/Tab-order evidence
- [ ] Live re-verification after this merges/deploys to staging — will follow up

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -1606,19 +1606,35 @@ let twoincDomHelper = {
       // `close()` BEFORE `destroy()` — new, #30.x.13. Reached here the
       // widget is essentially always still OPEN: this function only runs
       // from activating the manual-entry row, and that row exists only
-      // INSIDE the open results list. selectWoo's own close cleanup (the
-      // work that unbinds its document-level "close on outside click" and
-      // Tab/Enter-as-select handlers — see the long comment in
-      // bindManualEntryAffordance above, which already documented this
-      // exact class of gap as a "known, DELIBERATELY UNFIXED" risk) only
-      // runs when the widget is told to close; `destroy()` on an OPEN
-      // widget tears down the widget's own DOM without ever running that
-      // cleanup, leaving those document-level handlers bound to
-      // `document` indefinitely — with no live widget left for them to
+      // INSIDE the open results list.
+      //
+      // CORRECTED mechanism (round 2 review, Han+Vader, both independently
+      // verified directly against the real vendored selectWoo.full.js —
+      // the previous version of this comment had the mechanism wrong):
+      // selectWoo's document-level keydown handler (bound ONCE per widget
+      // instance in `_registerEvents`, `$(document).on('keydown', ...)`,
+      // see the long comment in bindManualEntryAffordance above) is never
+      // unbound by anything — not by `close()`, not by `destroy()`. What
+      // actually neutralizes it: that handler's dangerous branches are
+      // gated on `self.isOpen()`, which just reads a CSS class
+      // (`select2-container--open`) on the container. `close()`
+      // synchronously flips that class off. `destroy()` alone never fires
+      // the close event, so a destroyed-but-still-referenced instance's
+      // container keeps that class (and therefore `isOpen() === true`)
+      // forever — the handler is still bound to `document` and still
+      // "live" by its own gate, just with no widget left for it to
       // reason about. That is what live reproduction (#30.x.13, Doug)
       // showed: Tab became unresponsive PAGE-WIDE, not just near the
-      // company field, the moment manual entry was reached, exactly what a
-      // stray document-level Tab interceptor with no scoping would produce.
+      // company field, the moment manual entry was reached — exactly
+      // what that ungated zombie handler produces. Calling `close()`
+      // first is what actually fixes it, by flipping the one flag the
+      // handler checks; the handler itself is still bound afterward and
+      // always will be, so this is a mitigation of a permanent gap, not
+      // a removal of it. Direct empirical repro (round 2, Vader): a real
+      // widget destroyed WITHOUT close() first leaves a subsequent
+      // synthetic document keydown{which:9} reporting
+      // `defaultPrevented === true` (Tab trapped); with close() first,
+      // the same dispatch reports `false` (Tab free).
       //
       // Safe to call unconditionally ahead of destroy(): `close()` on an
       // already-closed widget is a documented no-op in select2/selectWoo.
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -2456,6 +2456,20 @@ let twoincSoleTrader = {
       window.twoinc.manual_company_entry_active = false;
       const $display = jQuery("#billing_company_display");
       if ($display.data("select2")) {
+        // close() before destroy() — same fix, same reason, as
+        // enterManualCompanyEntry (#30.x.13). This branch is reachable with
+        // the widget still OPEN exactly like that one: a buyer mid manual
+        // entry (dropdown already torn down, so a no-op there) or mid an
+        // open search (dropdown live) can switch to sole-trader mode via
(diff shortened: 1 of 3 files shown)


dgjlindsay:
## Adversarial review summary (self-review lane — reviewDecision not enforced on this repo)

**Round 1** — Han (SRE) + Vader (adversarial QA), parallel:
- Han: found `twoincSoleTrader.setMode`'s sole-trader branch had the identical destroy-without-close hazard as `enterManualCompanyEntry` (destroy() on an open selectWoo widget skips its own close cleanup). Fixed: close() before destroy() there too.
- Vader: found `manual_company_entry_active` was not saved/restored around a sole-trader-mode round trip, unlike `savedCompanySearch` — real reproducible dead end (buyer stuck on a required, unfillable Company ID field with no working search widget). Fixed: added `savedManualEntryActive`, saved/restored symmetrically.

Both fixes are mutation-verified (each new/changed test fails against the pre-fix code, passes against the fix — verified via `git stash`).

**Round 2** — Han + Vader again, read-only this time, against the updated diff:
- Vader: **CLEAN ROUND** — no new BLOCKER/BUG. Confirmed both round-1 fixes correct (sentinel logic, undefined-vs-false handling, no re-entrancy hazard). Two low-severity OBSERVATIONs, both pre-existing and unrelated to this diff's lines (a `preventDefault` micro-detail confirmed harmless; a pre-existing double-`selectWoo()`-init risk on country change, untouched by this PR).
- Han: confirmed both round-1 fixes correct and the two specific hazards asked about are handled properly. Found one additional **pre-existing bug**, not introduced by this PR: `setMode`'s business-restore branch (`if (window.twoinc.enable_company_search !== "yes") { getSearchCompanyBtnNode().show(); }`) only checks `enable_company_search`, not the new `manual_company_entry_active` flag — so a merchant with company search disabled entirely (not manual entry) who round-trips through sole-trader mode can end up shown a "search again" button that activates a search feature they never enabled. Recommended as a follow-up, not a merge blocker — narrow config (merchant-search-off + sole-trader-available), unrelated to the diff lines this PR touches.

**Verdict:** clean round reached. Not fixing the adjacent pre-existing gap in this PR — flagging to file as a follow-up ticket instead, to keep this PR's diff scoped to the 3 regressions it was opened for.

— by Claude

dgjlindsay:
## Round 2 adversarial review (fresh, independent — Han + Vader) + round 3 fix

The previous "round 2 clean" claim in this thread was **not** independently re-verified before now — re-ran both personas from scratch against this diff, without treating that claim as ground truth.

**Convergent BUG (both Han and Vader, independently):** `activateManualEntry` defers the actual manual-entry switch via `setTimeout(enterManualCompanyEntry, 0)`. The email-driven autofill prefetch can call `twoincSoleTrader.setMode("sole_trader")` asynchronously in that same window. If `setMode` wins the race, the stale deferred `enterManualCompanyEntry` still fires afterward and stomps `manual_company_entry_active` back to `true`, re-shows the search-again button `setMode` just hid, and wipes the fields sole-trader mode had just set — reproducing the exact #30.x.13 wrong-id-field-visibility symptom via a path this PR's own new flag opened up. **Fixed**: `enterManualCompanyEntry` now bails if sole-trader mode has taken over by the time the deferred callback runs, mirroring the existing "button removed before deferring" reentrancy guard. Mutation-verified (new test fails pre-fix, passes post-fix).

**NIT (both, independently, empirically verified against the real vendored `selectWoo.full.js`):** the code comment's claimed mechanism for the close()-before-destroy() fix was wrong — `close()` does not unbind selectWoo's document-level keydown handler (nothing ever does); it flips the `isOpen()` gate that handler's dangerous branches check, so the handler self-neuters rather than being removed. The *fix* (call order) was already correct — only the explanation was wrong. **Corrected** the comment. Vader additionally ran a direct empirical repro against the real library confirming the mechanism: destroy-only leaves a synthetic Tab keydown on `document` reporting `defaultPrevented === true` (trapped); close-then-destroy reports `false` (free) — this is the strongest evidence available that the Tab-trap fix (#30.x.13, the one flagged in this PR's own test plan as unconfirmed) actually works.

**Round 3 (after the above fix + doc correction):** all 209 jest tests green (206 baseline + 3 new: search-again detached-click regression, close-before-destroy order, and the new race-condition guard). No other findings from either reviewer against the button-binding change, the `manual_company_entry_active` null-sentinel snapshot/restore pattern, or the `toggleBusinessFields` branch split.

**Live browser verification — status, plainly:**
- Tab bug: live-tested on `woocom-dev.staging.two.inc` with real clicks/keypresses — clicked into an unrelated field then Tab, clicked into the manual-entry company field then Tab, typed then Tab — no trap in any case, focus moved normally each time (including landing on `#search_company_btn`, confirming that element exists in the DOM at that point).
- Company-ID-field-visibility and search-again-button fixes: attempted live verification on the same staging host and got the OLD (pre-fix) behavior in both cases (Company ID visible, search-again click doing nothing). On investigation this is very likely because `woocom-dev.staging.two.inc` git-syncs the `staging` branch, not this unmerged PR branch — this repo's own `deploy.yaml` has no PR-preview mechanism, so there's no way for that host to be running this diff yet. I'm not able to honestly claim live pre-merge confirmation of those two fixes on real deployed code — the unit tests are the evidence for those, mutation-verified against pre-fix code exactly as the PR description states. Real live re-verification of all three fixes (per the test plan's own open checkbox) has to happen post-merge, once `staging` actually carries this commit — flagging this clearly rather than rounding up to "live-verified."

**Verdict:** clean as of the round-3 fix + doc correction — no new findings from either reviewer against the current diff. CI green except `e2e-tests`, which is a pre-existing infra flake unrelated to this diff (same job fails identically on a direct run of `staging` itself — admin-login 404, order-flow test, nothing to do with the JS changed here).

— by Claude

PR #429, opened 2026-08-02
fix(company-search): a captured company must not lock the buyer out of manual entry or the keyboard (TWO-25326 §2/§4)

dgjlindsay:
Fixes two of the three defects Doug found live on 2026-08-02. **Both are verified in a real browser** — see the evidence section. The third is diagnosed and turns out not to be what it looked like; details at the bottom.

## The two defects were one regression

The capture gate added in PR #427 removed the "My company is not on the list" button as soon as `#billing_company_display` held a value. That produced both reports:

**§2 — "invisible entirely, worse than before".** Pick a company, reopen the dropdown to correct it, and there is no route into manual entry at all. Worse than the search-threshold gate it replaced, because typing no longer brought it back either. Confirmed in the browser: with a company selected, `#company_not_in_btn` does not exist in the DOM.

**§1/§4 — "Tab still doesn't work".** The search-field Tab handler keys on that button existing. With no button it did `return` and let the keystroke through to selectWoo's own document-level handler, which swallows it whole — `preventDefault` plus refocus the search field. Measured in that exact state: one real Tab press, focus stayed on `input.select2-search__field`, dropdown still in the DOM, `isOpen()` still true. A keyboard trap, which §4 forbids outright.

So PR #428's traversal was not the problem — it was never reached. Verified separately: with the button present, Tab has been landing on `#billing_address_1` correctly all along.

## Changes

**The capture gate is gone.** §2's "hidden once a company IS selected" cannot mean what it did. This button only exists inside the dropdown, so it is only ever on screen while the dropdown is open — and the dropdown being open *is* the buyer searching again. The clause is satisfied by the dropdown being shut; it does not extend to locking someone out mid-correction.

**Tab out of the query field no longer depends on the button.** If there is no button to shortcut to — a brand overlay dropping the affordance, or any future gating — it closes the dropdown and advances to the next real tab stop rather than falling through to a handler that traps focus.

**The shortcut confirms the button actually took focus.** `.focus()` on a hidden or mid-transition element silently no-ops per the HTML spec. That assumption is what shipped broken in the Tab-out handler and had to be fixed in #428; the same assumption was still sitting in this handler. If the button will not take focus, this path closes and advances too.

## Browser evidence

Live staging checkout, this build's code injected into the page, real mouse clicks and real Tab keypresses (not synthetic events):

| Step | Result |
|---|---|
| Company selected, dropdown reopened | `#company_not_in_btn` **exists**, `display: block`, `visibility: visible`, rect 602×41, offsetParent is the dropdown panel — where before the patch, in this exact state, it did not exist at all |
| Tab from the query field | focus → `BUTTON#company_not_in_btn`, dropdown still open |
| Tab again | focus → `INPUT#billing_address_1`, `select2-dropdown` gone from the DOM, `isOpen()` **false** |
| Same, from a dropdown opened without typing anything | identical: button, then `#billing_address_1`, `isOpen()` false |

## Tests

`npm run test:js` — 234 passed, 7 suites. `prettier@3.1.0` clean (pinned to the pre-commit version).

New coverage: the button survives a capture and a dropdown reopen; a captured company does not disable the Tab shortcut; Tab with no button still closes and advances; a button that refuses focus is not a dead end. The last one needs an element whose `focus()` is a no-op, modelled explicitly — jsdom does not enforce the spec's non-rendered rule.

## The third defect is not a label bug — do not "fix" it the way it was described

Reported as: the tile's name/number label appears on selection, then disappears once the intent check begins.

It did not reproduce on an idle intent check (polled every 500ms for 60s — label populated on selection and stayed populated straight through the intent box appearing, same DOM node throughout), nor on a forced `update_checkout`, nor on a term-chip click.

It **does** reproduce on a payment-method round trip — switch to another method and back. But the label is not being wiped. The captured company is:

```
{"probe":"P1","txt":"","n":1,
 "sum":"<div id=\"twoinc_company_summary\" class=\"twoinc-company-summary hidden\"><span class=\"twoinc-company-summary-id\"></span></div>",
 "bc":"","cid":""}
```

`data-probe` survived, so the element was never replaced — it was correctly re-rendered as empty, because `#billing_company` and `#company_id` are **both empty**. The address-area summary is blank for the same reason. Calling `renderCompanySummary()` by hand does not repopulate it, correctly, because there is nothing captured to render.

This diff and the two before it never write those two fields (checked: no such write appears anywhere in the added lines). The plausible mechanism is `clearSelectedCompany()` firing from `onCountryInputChange` on a spurious `change` during WooCommerce's checkout re-render — which is TWO-24867, "country-switch breaks company search and autofill (PS, WC)", already linked from TWO-25326.

Two reasons this matters more than a cosmetic label:

1. It is a data-integrity bug, not a display one. The buyer sees a company in the combobox while the fields the order actually posts are empty — §6 territory.
2. **Making the label persist here would be the wrong fix.** It would display a company the order is not carrying, which is precisely the failure the existing `readCapturedCompany` comments were written to prevent. The label is doing its job by going blank; it is the first thing in the tile that ever made this visible.

Left for a decision rather than patched over. Nothing in this PR changes that behaviour either way.

## Other

Leak gate clean for this diff. One incidental, not chased and not introduced here: on a fresh load the first real click on the company control sometimes does not open the dropdown and a second is needed — seen intermittently across several browser runs, and it may be a bridge click-dispatch artefact rather than a page bug.

https://claude.ai/code/session_01XHqsP6xutJexRoEDybQoV4

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -284,20 +284,29 @@ let twoincSelectWooHelper = {
   syncManualEntryButton: function () {
     const helper = twoincSelectWooHelper;
 
-    const $display = jQuery("#billing_company_display");
-    const picker = $display.data("select2");
+    const picker = jQuery("#billing_company_display").data("select2");
     if (!picker || !picker.$results || !picker.$results.length) return;
 
     const $list = picker.$results;
     const $existing = jQuery("#" + helper.manualEntryRowId);
 
-    // The empty option's label is a non-breaking space, so an unselected
-    // picker reads back as " " rather than "" — the same trap
-    // renderCompanySummary documents. blankToEmpty collapses both.
-    if (twoincUtilHelper.blankToEmpty($display.val())) {
-      $existing.remove();
-      return;
-    }
+    // No capture gate. There was one — the button was removed as soon as
+    // `#billing_company_display` held a value — and it was wrong, found live
+    // by Doug 2026-08-02: pick a company, reopen the dropdown to change it,
+    // and there was no route into manual entry at all any more. Worse than
+    // the threshold gate it replaced, because typing no longer brought it
+    // back either.
+    //
+    // §2's "hidden once a company IS selected" cannot mean that. This button
+    // only ever exists inside the dropdown, so it is only ever on screen
+    // while the dropdown is open — and the dropdown being open IS the buyer
+    // searching again. "Hidden once selected" is satisfied by the dropdown
+    // being shut; it does not extend to locking a buyer out of manual entry
+    // because of a pick they are in the middle of correcting.
+    //
+    // That gate also silently disabled the Tab shortcut, since the handler
+    // below keys on the button existing — which is how one regression
+    // presented as two (§2 invisible, §4 keyboard trap).
 
     // Already there, immediately after the current results list: nothing to
     // do. Load-bearing rather than an optimisation for the same reason it was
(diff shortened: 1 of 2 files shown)


dgjlindsay:
Self-review summary (self-review lane — `reviewDecision` is unenforced on this repo).

**Adversarial pass over the diff, findings addressed before pushing:**

1. *Removing the capture gate could resurrect the pre-#427 defect Doug originally filed (§2 "must not require typing a doomed query first").* Checked: it does not. The `select2:open` placement added in #427 is untouched, so the button is still present the moment the dropdown opens with nothing typed — verified live in this run, not just reasoned about.
2. *Does dropping the gate leave the button visible when it should not be?* It only ever exists inside the dropdown, and selectWoo detaches the whole dropdown container on close, so there is no state where it is on screen with the dropdown shut.
3. *The no-button Tab branch duplicates the Tab-out handler.* Deliberate — repetition over a shared helper for two handlers with different interferers, matching the existing convention in this file (see the note contrasting the return-to-search button's Enter/Space fix with this one).
4. *`.focus()`-assumed-to-work was fixed in the Tab-out handler in #428 but the identical assumption was still live in the shortcut handler.* Now fixed in both. This is the third time this specific assumption has caused a defect on this ticket.

**Regression risk considered:** nothing here writes `#billing_company` or `#company_id`, and no existing test changed meaning except the two that asserted the removed gate — both rewritten to assert the corrected behaviour with the reasoning inline. 234 JS tests, the PHP suite, and prettier are all green; the only red check is `e2e-tests`, which has been failing on `staging` since 2026-07-31 with `admin login failed: 404` (pre-existing infra, unrelated to this change).

**Verification standard applied:** every claim above about behaviour was checked in a real browser on the staging checkout with real clicks and real Tab keypresses, with this build injected into the page. Two prior rounds on this ticket were reported fixed on static reasoning and were not.

PR #430, opened 2026-08-02
fix(checkout): react to a real country change, and only to a real one (TWO-24867)

dgjlindsay:
## TWO-24867 — country switch breaks company search & autofill (PrestaShop → WooCommerce)

Ports the PrestaShop country-switch fix to WooCommerce, and fixes the
spurious-clear defect found live while working TWO-25326 — which turns out to
be the same bug class pointing the other way.

Everything here is one sentence: **the country-sensitive code and the billing
country field could disagree**, in four different ways.

### 1. The handler fired when the country had not changed (TWO-25326)

`onCountryInputChange` is bound delegated on `document.body` and acted on every
`change` event `#billing_country` emitted. WooCommerce emits plenty that are not
a buyer changing country — `updated_checkout` re-renders the billing fields, and
core's `address-i18n.js` re-triggers the field on `country_to_state_changing` at
checkout init. Each one ran `clearSelectedCompany()` and destroyed the captured
company (`#billing_company`, `#company_id`, the picker's selection, the registry
address behind it) with the buyer having done nothing.

Now guarded on the value actually differing from the last country the page acted
on. Compared by value rather than by testing `event.originalEvent`: WooCommerce's
re-render path re-triggers through jQuery on some themes and dispatches a native
event on others, so an event-source test would hold on the fixture and fail on
the shop.

**Only the destructive half is behind the guard.** `toggleBusinessFields()` stays
unconditional — the swallowed events are precisely the ones that just re-rendered
the billing fields underneath it, so gating the visibility pass too would have
turned this fix into a field-visibility regression.

### 2. The tracker could drift from the field

The guard is only as good as what it compares against, so `countryDidChange` is
the single writer and every path goes through it.

**The seed was taken too early — this was the blocker.** `initialize()` seeded
next to the binding that reads it, then went on to call `loadStorageInputs()`,
which writes `#billing_country` with `selectElem.value =` and fires no `change`.
So the tracker held the country the page was *rendered* with while the field held
the *restored* one, and the first re-render afterwards read the difference as a
real country change — destroying the company and address that same restore had
just put back. `initialize(true)` is the bootstrap's own call, so this was the
production path for any buyer returning to a checkout mid-session. Seeded at the
end of `initialize()` now, after both restore passes.

**`updated_checkout` RECORDS the country, and deliberately does no more.** A
`change` is not guaranteed: WooCommerce can replace the billing fields with a
server-rendered country and fire nothing — a `checkout_error` re-render, a
multi-step theme, a session address restored server-side. Without recording,
the tracker kept the pre-re-render country for the rest of the page and a later
genuine switch *back* to it read as no change and was swallowed.

It must not do more than record. Those same re-renders restore the country and
the company **together**, so clearing the capture there destroys what the
re-render just put back — the same failure this PR exists to remove, on a new
trigger. Throwing away a captured company needs the buyer's gesture, and a
`change` event is the only signal of one this checkout has.

Two readings are deliberately not a change: an **empty** one (the field
mid-replacement) — not acted on and *not recorded*, so a switch completing after
the gap is still compared against the last real country — and the **first known**
country, where there is nothing to have moved away from.

### 3. The search queried the country the widget was built under

`genSelectWooParams()` closed over the country. That held only while *every*
country change went through `clearSelectedCompany()`, which rebuilds the widget —
and it no longer does: the guard above deliberately skips the re-render events,
and a country written programmatically (`.val()`) fires no `change` at all. Both
leave the widget alive with a stale closure, so the buyer got "no companies
found" for a company that exists. Read per request now.

This inverts a test that pinned the old behaviour deliberately
(`company-search-results.test.js`); updated with the reasoning rather than deleted.

### 4. The registry address lookup had no supersession guard

A response for a company picked under the outgoing country wrote its address over
the fields after the switch, and set `registryAddressApplied` on it. Guarded
twice, because neither subsumes the other:

- a sequence number — bumped by a newer lookup **and** by a real country change
  (the only thing covering a switch away and back, where the country matches
  again by the time the response lands);
- the country snapshot taken when the request was issued — the only thing
  covering a country moved programmatically, where nothing bumps the sequence.

An empty reading on **either** side is not a mismatch: discarding a good registry
address would be a silent failure with no retry and no message, and a lookup
issued while WooCommerce was mid-replacement snapshots `""`, which compared
against a known country would discard every response for it.

`twoincSoleTrader.currentCountry()` also delegates to the shared reader now
rather than keeping its own copy of the same two lines — its per-country
availability cache would otherwise be keyed on one answer and read with another.

### Plus

`customerCompany.country_prefix` is written **after** `clearSelectedCompany()`
rather than before it. That function resets the object wholesale and only re-reads
it from the DOM three seconds later, so the assignment was dropped and
`getApproval()` — plus `getDueInDays()`, which early-returns without one — ran
with no country for that window.

### Acceptance criteria

- *Switching country re-runs company search and re-populates autofill with no
  stale state* — (3) and (4).
- *No 500 from stale formatter state* — the PrestaShop 500 came from an address
  formatter override that repo carries; **WooCommerce has no equivalent**. The one
  piece of server-side per-country state here,
  `WC_Twoinc_Sole_Trader::$types_cache`, is already keyed by ISO country and
  session-scoped — checked as part of this work. Nothing to port; recorded so the
  gap is not re-investigated.
- *Behaviour consistent across platforms* — Magento remains outstanding on the ticket.

### Tests

`tests/js/country-switch.test.js`, 19 new tests. Suite: **257 passed**.

The whole suite drives the **real** wiring — a real `initialize(false)`, then
`trigger("change")` on the field — so unwiring the delegated binding, or moving
the seed out of `initialize()`'s reach, fails it. That restructure exposed a leak
worth naming: those bindings are delegated on `document.body`, which survives
`innerHTML = ""`, so each test was leaving a live handler closed over its own
evaluation of the source and the next test's event ran all of them. Unbound in
`afterEach`; it had been producing an order-dependent pass.

Mutation-checked — each fix reverted individually, the named test confirmed red,
then restored green:

| mutation | caught by |
|---|---|
| guard removed (handler always acts) | 6 tests |
| `toggleBusinessFields` moved behind the guard | "still re-runs the field-visibility pass" |
| `initialize()` seed removed | 9 tests |
| empty-reading early return removed | "an empty reading is neither acted on nor recorded" |
| first-known-adopt removed | "the FIRST country ever seen is adopted, not acted on" |
| `countryDidChange` stops recording | 8 tests |
| seed moved back before the restore passes | "the seed is taken AFTER the saved-input restore, not before" |
| `onUpdatedCheckout` recording removed | 2 tests |
| `onUpdatedCheckout` recording made destructive | "updated_checkout records WITHOUT clearing the captured company" |
| closure capture restored | live-country test + the inverted results test |
| search-seq bump removed | 2 tests |
| `addressLookupSeq` bump removed | "a response is discarded after a switch AWAY and BACK" |
| lookup sequence guard removed | "a response superseded by a newer lookup is discarded" |
| lookup country guard removed | "a response is discarded when the country moved on WITHOUT an event" |
| `country_prefix` set before the clear | "leaves the new country on customerCompany" |

**One line is deliberately uncovered** and says so in its own comment: the spinner
hand-off in the country handler. Nothing reachable fails without it — what clears
the spinner today is `clearSelectedCompany()` re-attaching the widget and taking
the dropdown with it — so a test would pass either way, which is worse than none.

PHP unit suite green. `tests/js/README.md` updated for the new suite and for the
inverted claim.

<sub>PR description by Claude.</sub>

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -936,6 +936,9 @@ let twoincSelectWooHelper = {
    * supersession check — so those three can never disagree about what "the
    * current country" is.
    *
+   * `twoincSoleTrader.currentCountry()` delegates to this one, so the
+   * per-country availability cache cannot be keyed on a different answer.
+   *
    * NOT the only reader in the file: `getCompanyData()` and
    * `isCountrySupported()` still read `.val()` raw and uncased, so the
    * `country_prefix` this file's handler writes upper-cased is replaced with
@@ -2895,8 +2898,12 @@ let twoincSoleTrader = {
     return (window.twoinc && window.twoinc.sole_trader) || {};
   },
 
+  // Delegated rather than a second copy of the same two lines (TWO-24867):
+  // sole-trader availability is decided per country and cached per country,
+  // so it disagreeing with the country the search and the change guard use
+  // would be a cache keyed on one answer and read with another.
   currentCountry: function () {
-    return (jQuery("#billing_country").val() || "").toUpperCase();
+    return twoincSelectWooHelper.currentCountry();
   },
 
   enteredEmail: function () {
@@ -3561,18 +3568,8 @@ class Twoinc {
       twoincDomHelper.togglePaySubtitleDesc();
     });
 
-    // Handle the country inputs change event.
-    //
-    // Seeded first (TWO-25326 / TWO-24867), through the same single writer
-    // the handler uses. The handler only acts when the value differs from the
-    // last one recorded, and with no seed the FIRST country the page ever
-    // sees is adopted rather than acted on — correct for the re-render event
-    // WooCommerce fires at init (core's address-i18n.js triggers
-    // `country_to_state_changing` carrying the country the form already had),
-    // but wrong for a buyer who changes country before any re-render happens:
-    // their switch would be the first known country and would be swallowed.
-    // Seeding here is what tells those two apart.
-    twoincSelectWooHelper.countryDidChange(twoincSelectWooHelper.currentCountry());
+    // Handle the country inputs change event. The tracker behind it is seeded
+    // at the END of this function, not here — see the comment there.
     $body.on("change", "#billing_country", self.onCountryInputChange);
 
     // Re-evaluate the sole-trader autofill prefetch whenever the email
@@ -3604,6 +3601,29 @@ class Twoinc {
 
     twoincDomHelper.loadUserMetaInputs();
     if (loadSavedInputs) twoincDomHelper.loadStorageInputs();
+
+    // Seed the country tracker HERE — after the two restore passes above, not
+    // next to the binding that reads it (TWO-24867 / TWO-25326).
+    //
+    // `loadStorageInputs()` writes #billing_country with `selectElem.value =`
+    // and fires no `change`. Seeded before it, the tracker held the country
+    // the page was rendered with while the field held the restored one, and
+    // the first re-render afterwards read the difference as a real country
+    // change — destroying the company and address that same restore had just
+    // put back. The bootstrap's own call is `initialize(true)`, so that is the
+    // production path, not an edge case.
+    //
+    // Seeding at all is what tells the two first-event cases apart: with no
+    // seed the FIRST country the page ever sees is adopted rather than acted
+    // on — right for the re-render WooCommerce fires at init (core's
+    // address-i18n.js triggers `country_to_state_changing` carrying the
+    // country the form already had), wrong for a buyer who changes country
+    // before any re-render happens.
+    //
+    // Through `countryDidChange` rather than by assignment, so this file has
+    // exactly one writer for the tracker.
+    twoincSelectWooHelper.countryDidChange(twoincSelectWooHelper.currentCountry());
+
     setTimeout(function () {
       twoincDomHelper.saveCheckoutInputs();
       Twoinc.getInstance().customerCompany = twoincDomHelper.getCompanyData();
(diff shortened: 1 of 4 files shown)


dgjlindsay:
## Self-review summary — three adversarial rounds, not merged

Recorded here so the reasoning is on the PR rather than only in commit messages.

**Round 1** — two majors. The country tracker had one writer and never self-healed, so `initialize()`'s single seed could go stale; and an empty reading (`#billing_country` mid-replacement) read as a country change. Also: the `addressLookupSeq` bump and the `initialize()` seeding were both uncovered, and the tests drove the handler directly rather than through the delegated binding it is actually reached by. That last one mattered — fixing it exposed that delegated `document.body` handlers survive `innerHTML = ""`, so the suite had been passing order-dependently.

**Round 2** — a blocker, with a reproduction. The seed was taken before `loadStorageInputs()`, which writes `#billing_country` with `selectElem.value =` and fires no `change`. The tracker therefore held the country the page was *rendered* with while the field held the *restored* one, and the first re-render read the difference as a real country change — destroying the company and address that same restore had just put back. `initialize(true)` is the bootstrap's own call, so that was the production path for a buyer returning mid-session. Round 2 also moved `updated_checkout` from running the whole handler to recording only, because those re-renders restore the country and the company together.

**Round 3** — two minors, no blocker. One fixed (record-only left in-flight work for the outgoing country uninvalidated). Two deferred with reasons, in the final commit message rather than implicitly: a wrong-country company is still reachable when the country moves with no `change` event and nothing downstream catches it — checked, not assumed: `getApproval()` posts the old `country_prefix` so it self-consistently approves, and the order payload pairs the posted company id with the order's billing country with no consistency check, so it surfaces as an opaque order-creation failure. The suggested discriminator looks right but is a behaviour change on the destructive path and wants its own ticket and review round. The other is pre-existing (`initialize()` early-returns without setting `isInitialized`).

### Why this is not being merged

Round 3's findings were consequences of round 2's fix. That is an oscillating loop, and the rule is to stop and report rather than stack a fourth patch — so this wants a human call.

### Verification status, stated precisely

- **The bug is live-verified. The fix is not.** On the dev twin serving `staging`, a real GB company was captured and then a `change` was fired on `#billing_country` with no value change (GB → GB): `#billing_company`, `#company_id`, `billing_address_1`, `billing_city` and `billing_postcode` were all wiped to `""`. An `update_checkout` trigger alone wiped nothing, so the trigger is the country field's own event, not the re-render in general.
- The fix could not be exercised in a browser: no shop here serves an unmerged branch, and the Chrome extension dropped before the branch build could be injected into the live page. Merging is what makes live verification possible, which is worth weighing in the merge call.
- Jest **259 passing**, 25 new tests. **17 mutations** caught, each reverted individually with the named test confirmed red then restored. An eighteenth survived — the spinner hand-off — and that line is marked in its own comment as deliberately uncovered rather than counted as covered.
- PHP unit suite green. `e2e-tests` is red on **every** branch in this repo including merged ones (`admin login failed: 404`, a harness failure on the shared base); it is not caused by this diff.

<sub>Review and write-up by Claude.</sub>

PR #431, opened 2026-08-03
fix(company-search): tie the tile company label to the intent message's visibility (TWO-25326 §7)

dgjlindsay:
## What

TWO-25326 §7, revised 2026-08-03. The payment tile's captured-company label is now shown **exactly when the intent-approved notice is shown**, and hidden exactly when it is hidden.

This supersedes the rule that shipped in woocommerce-plugin PR #429, which showed the label whenever a company had been captured, regardless of anything else.

## How

**One gate, not two.** A new `isIntentNoticeVisible()` reads the notice *element's* own state back out of the DOM; `renderCompanyTileLabel()` toggles the label on that. The label follows the notice rather than re-deriving the condition that produced it — a parallel copy of the condition would agree today and drift later, which is the specific failure this shape rules out.

Both halves of that read are load-bearing, covering the two independent ways the notice can be off screen:

- **absent** — the brand set `intent_approved_notice_enabled` false, so `get_intent_approved_notice()` returned `''` and the div was never rendered;
- **`hidden`** — the notice exists but the checkout is not in the approved state (pre-intent, checking, errored, mid-re-render).

**Reactive, not render-once.** `togglePaySubtitleDesc` is the only thing in the plugin that changes the notice's visibility, so the label re-sync sits at the end of it, after the notice's class has settled. It calls the label renderer with no arguments, which reads the captured company back off the live inputs — so the capture sites don't have to know the notice moved. Without this the label would only ever be correct on whichever render happened to follow a capture, and e.g. an error state would leave it stranded on screen.

Two supporting changes:

- **The label container is no longer emitted** when the brand has the notice switched off. With the notice suppressed there is no state in which the label can appear, so the markup would only ever sit empty and hidden. Same reasoning TWO-25224 already applied to the intent loader. The JS gate does not depend on this — it holds on its own — the two agree because they are the same rule.
- **`.twoinc-company-tile-label:empty { display: none }`.** The JS gate deliberately does *not* and-in "has text", so that "shown when the notice is shown" is literally true of that one line. An intent-approved checkout with no known company is a real state — the notice ships a separate no-company sentence precisely for it — and would otherwise leave an empty label holding its own `12px` margins open: 24px of gap between the chips and the notice, for nothing. The no-content case belongs in CSS, not in the visibility policy.

## Tests

8 new JS cases in the new `tile label visibility mirrors the intent notice` block, asserting the label's hidden state against **the notice's**, never against the intent action — a test keyed on the action would pass just as happily against the parallel re-derivation this is built to avoid. Includes the both-directions sweep over the full action set (with an explicit check that the sweep visited both states, so the invariant can't be satisfied by everything staying hidden), the brand-suppressed case, the re-sync case, and the no-company case.

1 new PHP case pins the container to the notice switch in **both** directions — presence alone would pass against unconditional markup, absence alone against markup that never emits it.

The pre-existing content assertions now drive the tile into the approved state first, since that is the only state the label is on screen in; the notice is part of the test tile's minimum markup as a result. Where a capture or a re-render legitimately invalidates the prior approval, the test re-approves explicitly with a comment saying so — that hiding is the new behaviour, not a harness quirk.

### Verification

| Check | Result |
|---|---|
| `npm run test:js` | 242 passed (was 234; +8) |
| `php tests/unit/run.php` | all passed (+1 new case) |
| `pre-commit` (prettier) | passed |
| Mutation: gate reverted to old captured-company rule | **fails 5 cases** |
| Mutation: re-sync call removed | **fails 7 cases** |
| Mutation: absent-notice half of the gate dropped | **fails 1 case** |
| Mutation: `:empty` CSS rule removed | **fails 1 case** |
| Mutation: container always emitted | **fails the PHP case** |

Every mutation was confirmed to have actually applied to the tree before its verdict was taken.

**Not verified in a real browser.** The Chrome bridge was disconnecting repeatedly during this session. Verification here is unit + mutation + CI only. The reactive path in particular (notice hides → label hides in the same turn) is asserted in jsdom against the real `togglePaySubtitleDesc`, but has not been watched happen on a live checkout.

`phpcs` and `phpstan` are not installed locally; CI covers both.

---

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -617,6 +617,24 @@
 .twoinc-company-tile-label.hidden {
   display: none;
 }
+/*
+ * The no-content case (TWO-25326 §7, revised 2026-08-03).
+ *
+ * The label's visibility now mirrors the intent-approved notice's, and NOT
+ * "has a company been captured" — so the label can be unhidden while holding
+ * no text, on an intent-approved checkout whose company is unknown. That is a
+ * real state rather than a defensive hypothetical: the notice itself carries a
+ * separate no-company variant of its sentence for exactly it.
+ *
+ * Without this rule such a label would be an invisible 24px gap (the 12px
+ * margins above and below) wedged between the chips and the notice. Handled
+ * here rather than by and-ing the text back into the JS condition, so the one
+ * gate in twoinc.js stays literally "shown when the notice is shown" and this
+ * stays what it is — a rendering detail about an element with nothing in it.
+ */
+.twoinc-company-tile-label:empty {
+  display: none;
+}
 .twoinc-mode-selector {
   display: flex;
   flex-wrap: wrap;
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -1580,6 +1580,47 @@ let twoincDomHelper = {
         jQuery(".twoinc-pay-box" + errSelector).removeClass("hidden");
       }
     }
+
+    // The tile's captured-company label tracks the intent notice's
+    // visibility (TWO-25326 §7, revised 2026-08-03), and this method is the
+    // ONLY thing in the plugin that changes that visibility — the sweep on
+    // the first line plus the three branches above are the whole set. So the
+    // re-sync belongs here, at the end, AFTER the notice's own class has
+    // settled: `renderCompanyTileLabel()` reads that class back rather than
+    // recomputing the action, which is what keeps the two from drifting.
+    //
+    // Called unconditionally, including for the actions that show nothing:
+    // hiding the notice must hide the label in the same turn, and the early
+    // `addClass` sweep means every non-intent action is a hide.
+    twoincDomHelper.renderCompanyTileLabel();
+  },
+
+  /**
+   * Whether the payment tile's intent-approved notice is on screen right now
+   * (TWO-25326 §7, revised 2026-08-03).
+   *
+   * The single source of truth for the captured-company label's visibility.
+   * Deliberately reads the notice ELEMENT's own state rather than re-deriving
+   * it from the intent action: the requirement is that the label shows
+   * exactly when the notice shows, and any second copy of the condition — however
+   * faithful when written — is a thing that can come to disagree.
+   *
+   * Both halves are load-bearing, and they cover the two independent ways the
+   * notice can be off screen:
+   *   - `.length` — the brand switched the notice off, so
+   *     get_intent_approved_notice() returned '' and the div was never
+   *     rendered at all (`intent_approved_notice_enabled: false`). An absent
+   *     notice is a hidden notice, so the label stays hidden forever on that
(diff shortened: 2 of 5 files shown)


dgjlindsay:
## Self-review — 2 rounds, clean

Reviewed adversarially against the revised §7 rule: the label must be shown **exactly** when the intent-approved notice is shown, via the *same* gate rather than a parallel condition, and must react at runtime rather than only on render.

### Round 1 — one finding, fixed

**The no-company case was incompletely asserted.** It checked the notice was shown, the label's text was empty, and the `:empty` CSS rule existed — but never that the label element is left *unhidden*. That omission mattered: the test would have passed unchanged against an implementation that quietly and-ed `!text` back into the JS gate, at which point the CSS assertion beneath it would be guarding a state that could no longer arise.

Fixed in 57e2e63 with `expect(tileLabelShown()).toBe(true)`. Mutation-confirmed: and-ing `!text` back into the gate now fails that case, and failed nothing before it was added.

### Round 2 — clean, no new findings

Round 1 left one claim unverified — that `togglePaySubtitleDesc` really is the only thing that changes the notice's visibility, which the whole "mirror the element" approach rests on. Verified by grep this round rather than taken from the comment:

- `.twoinc-intent-approved` is mutated only inside `togglePaySubtitleDesc` (the blanket `.twoinc-pay-box` hide, and the one `removeClass` on the approved branch). The only other reference is the new helper, which reads.
- No broad visibility sweep can catch it. The one nearby candidate, `visibleTargets.removeClass("hidden")`, operates on billing *field* selectors only.
- Both call paths of `renderCompanyTileLabel` — with explicit args from `renderCompanySummary`, and with none from `togglePaySubtitleDesc` — apply the gate unconditionally, so neither bypasses it.
- Pay-for-order (`custom-checkout`) renders no payment tile at all: no notice *and* no label container, so there is no surface where one exists without the other.
- All 9 new cases were checked for tautology; each fails against a real defect.

### Verification status

Green: 242 Jest cases, the PHP suite, phpcs, phpstan, pre-commit, unit-tests across PHP 7.4–8.4, both boot smokes, upgrade smoke. Six mutations were applied and each was confirmed to have actually landed in the tree before its verdict was taken.

**`e2e-tests` is red, and it is not this branch.** It fails with `admin login failed: 404` from the staging shop's admin endpoint, and it is failing the same way on unrelated branches (`doug/TWO-24867-wc-country-switch`, repeatedly, since yesterday evening). Shared environment, not this diff — flagged rather than worked around.

**No live browser verification.** The reactive path is asserted in jsdom against the real `togglePaySubtitleDesc`, but nobody has watched the label follow the notice on a real checkout: the Chrome bridge was disconnecting through this session, and the e2e route needs merchant credentials that were not available. Worth a live pass on staging once deployed, particularly the transition into an error state with a company already captured.

---
<sub>Review by Claude.</sub>

PR #432, opened 2026-08-03
chore(TWO-25326): remove partner-initial ticket prefixes from comments

dgjlindsay:
## What

Comment-only cleanup. Eleven places in this **public** repo carried an
external ticket prefix whose letters are a partner's initials:

| File | Kind |
|---|---|
| `assets/css/twoinc.css` | CSS comment |
| `assets/js/twoinc.js` | JS comment |
| `class/WC_Twoinc.php` (x2) | PHP comments |
| `class/WC_Twoinc_Checkout.php` | PHP comment |
| `class/WC_Twoinc_Payment_Terms.php` | docblock |
| `tests/js/README.md` | prose |
| `tests/js/term-chips.test.js` | docblock |
| `tests/unit/run.php` (x3) | docblock + comment |

Each reference is dropped rather than reworded — the surrounding prose
already explains the behaviour it was pointing at, so nothing
load-bearing is lost. Two comments were re-wrapped to keep line lengths
tidy after the parenthetical came out.

## Not in scope

Old commits still contain the references. Per the task owner's
instruction, history is **not** rewritten — only the branch tip is made
clean.

## Verification

- `php tests/unit/run.php` — all tests passed
- `npm run test:js` — 7 suites, 242 tests passed
- `php -l` clean on all four touched PHP files
- Repo-wide case-insensitive grep for the partner name and the prefix
  returns zero hits at this tip (one base64 substring in
  `package-lock.json` is a false positive and untouched)
- Three-dot diff against `staging` contains the prefix only on removal
  (`-`) lines

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -403,8 +403,8 @@
 
 /* Payment terms chip selector (TWO-24751) */
 /*
- * Heading above the chips, shown only when more than one term is offered
- * (ABN-468). Mirrors Magento's <label class="label"> in the Luma
+ * Heading above the chips, shown only when more than one term is offered.
+ * Mirrors Magento's <label class="label"> in the Luma
  * gateway_method.html template — a plain form label, not a chip.
  */
 .twoinc-term-chips-heading {
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -3033,7 +3033,7 @@ let twoincTermChips = {
     // Heading placement mirrors Magento's Luma template: shown ABOVE the
     // chips only when the buyer has a choice to make. A single chip carries
     // its own "Payment Terms N days" label instead, so a heading there would
-    // say the same thing twice (ABN-468).
+    // say the same thing twice.
     const $heading = jQuery(".twoinc-term-chips-heading");
     if (single || terms.length === 0) {
       $heading.addClass("hidden").text("");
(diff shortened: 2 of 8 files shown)


dgjlindsay:
## Self-review — round 1: clean

Adversarial review of `origin/staging...HEAD`, one round, **no findings**. Checked:

1. **Prose integrity** — every comment still reads correctly with the reference removed; the
   surrounding sentences already carried the explanation. Two comments re-wrapped so line
   lengths stay tidy.
2. **No code touched** — all 14 changed lines are inside `/* */`, `//`, `/** */` or markdown prose.
   `git diff` confirms nothing outside a comment moved.
3. **No test asserted the removed literal** — the redacted tokens appear only in docblocks and
   inline comments, never in an assertion string.
4. **Residual scan** — repo-wide case-insensitive grep for the redacted prefix and the full name
   it abbreviates returns zero hits at this tip. The single `package-lock.json` match is a
   base64 integrity hash substring, not a name.

### Checks

Green: phpcs, phpstan, pre-commit, Jest (Node 20), unit-tests on PHP 7.4/8.0/8.1/8.2/8.3/8.4,
boot smoke on WooCommerce 10.8.1 and 10.9.4, gettext catalogues, upgrade smoke, version
computation.

`e2e-tests` is **red, and pre-existing** — it fails on every branch in the repo right now,
including `staging` itself and the last PR merged into it. The failure is
`admin login failed: 404` from `tests/e2e/two-api.ts:80` in `triggerFulfilBatch`, i.e. a remote
environment returning a not-found page, which a comment-only diff cannot cause or fix. Flagged
rather than fixed: it lives on the shared base, not in this change.

### Scope note

Only the branch tip is cleaned. Earlier commits still contain the references and are
deliberately left alone.

---

PR #433, opened 2026-08-03
fix(checkout): drop a company stranded in the wrong country (TWO-25333)

dgjlindsay:
## The gap

PR #430 (TWO-24867) made `onUpdatedCheckout` **record** a billing country that moved with no `change` event, and deliberately **not** clear the captured company — right for the case it was made for, because a `checkout_error` re-render, a multi-step theme or a server-side session restore supplies the country and the company *together*, and clearing there destroys what the same re-render just put back.

It left one case wrong. When the country really did move to something the captured company does not belong to, that company survived and nothing downstream caught it:

- `getApproval()` posts `customerCompany` carrying the **old** `country_prefix` next to the **old** organisation number, so the pair is internally consistent and the intent check approves it. The buyer sees a green payment method.
- The order payload pairs that `company_id` with the **order's** billing country, with no consistency check between the two.

So the mismatch was not caught at checkout. It reached order creation and came back as an opaque failure the buyer could not act on and the merchant could not diagnose from the storefront.

## The fix

Discriminate, rather than choosing between "always clear" and "never clear". On the record-only path, clear the capture **only when the recorded country and the captured company's own `country_prefix` actually disagree**.

This is specifically not a return of the TWO-25326 defect: in the restore-together case those two values agree by construction, because the same re-render supplied both, so the guard stays silent exactly where clearing would be destructive and fires only where the capture is genuinely stale. When a restore has a genuine mismatch — a stored company from a different country — clearing it is the correct behaviour, not a regression.

Three readings are deliberately **not** grounds to clear, each with its own test:

- **A company name with no organisation number.** Not a capture (TWO-25326 §6: the method is usable only for a company captured *with* an id), and nothing about a bare name is invalidated by a country change — the buyer typed it.
- **An unknown country on either side.** `country_prefix` is null until the first capture, and an empty field reading means the field was mid-replacement. Same rule the address-lookup guard and `countryDidChange` already apply.
- **The DOM already holding a *different* company from the one recorded.** `customerCompany` is refreshed from the DOM on a timer, so a re-render that swapped in another saved address — country *and* company together, a different pair but a self-consistent one — arrives with the previous capture still in JS state. There the record is stale rather than the fields, so it is re-synced to the DOM instead of destroying a company the re-render had just restored.

The comparison is case-insensitive, because the two sides are written by different readers: `currentCountry()` upper-cases and `getCompanyData()` reads the field raw (the inconsistency flagged in a comment on #430 and still not swept up). A false positive here is a destructive clear, so that is guarded rather than left to the observation that WooCommerce's values happen to be upper-case.

## Making the witness reliable

The ticket asked whether `country_prefix` is a reliable enough witness on every path. It was not, so this also **pins the capture country at all three capture sites**: the picker's `select2:select` handler, a manually typed organisation number, and the sole-trader setter (not on the clearing call `setMode("business")` makes with both arguments falsy).

Without that, the pair is assembled from two different moments — the number at capture, `country_prefix` from whichever DOM re-read ran last — so a country that moved with no `change` event *before* the capture left the old country next to a company from the new one. That both mis-posts the country on the intent and hands the discriminator a witness that can produce a false positive. Not scope creep: the discriminator is unsound without it.

## Verification

- **Unit: 283/283 pass** (`npm run test:js`, 8 suites), 41 in `country-switch.test.js` — 16 of them new.
- **Mutation-checked, 10 mutations, 0 survivors**, each caught by the intended test. Both directions the ticket demands are caught by disjoint sets: reverting the discriminator fails the stale-capture tests, widening it to "always clear" fails `updated_checkout records WITHOUT clearing the captured company`, `does NOT clear when the restore supplies country and company together`, `does NOT clear when the captured country is unknown` and both address-lookup variants.
- Also mutation-covered: the case normalisation, the DOM-differs guard, the no-capture guard, the post-clear country assignment, and each of the three capture-site pins (including making the sole-trader pin unconditional).
- Neither behaviour depends on `enable_address_lookup` — pinned on both settings, and the clear→`setAddress()`→`update_checkout` loop is pinned again for this path, which can now reach `clearSelectedCompany()`.
- **Not** browser-verified. The Chrome bridge is disconnected (checked, returns no connected extension instance), so this is unit and mutation evidence only.

## Note on a line removed rather than kept

A defensive supersession bump was written into the new helper and then removed: the only caller already bumps both counters unconditionally before reaching it, so the repeat changed nothing observable and no test could hold it in place. An untestable line that reads as the guarantee is worse than the guarantee living plainly at the call site.

## Docs

`tests/js/README.md` updated — the bullet stating that `updated_checkout` records "without clearing the capture" described behaviour this PR narrows.

Closes TWO-25333.

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -4070,9 +4070,12 @@ class Twoinc {
       twoincSelectWooHelper.companySearchSeq += 1;
       Twoinc.getInstance().addressLookupSeq += 1;
 
-      // BEFORE updateElements() below, which is what re-runs getApproval():
-      // called after it, the intent request for the stale pair would already
-      // have gone out and its answer would still be on screen.
+      // BEFORE updateElements() below, which is what re-runs getApproval().
+      // getApproval() does not fire immediately — it arms a 1s interval — so
+      // the ordering is not what stops the stale pair being posted, and no
+      // test pins it as though it were. It is here because clearing before the
+      // approval pass is the only order in which `updateElements` sees the
+      // state that the rest of this event's work should be derived from.
       Twoinc.getInstance().clearCompanyIfCountryStale(movedCountry);
     }
 
@@ -4100,13 +4103,23 @@ class Twoinc {
     let inputName = $input.attr("name");
 
     if (inputName === "company_id") {
-      Twoinc.getInstance().customerCompany.organization_number = $input.val();
-      // Pin the capture country next to the number the buyer just typed
-      // (TWO-25333 — see the picker's select handler for why the two have to
-      // be written together). Unconditional on the value: an emptied field
-      // leaves no capture for the country to be wrong about, and re-pinning it
-      // to the country now in the form is true either way.
-      Twoinc.getInstance().customerCompany.country_prefix = twoincSelectWooHelper.currentCountry();
+      const typed = $input.val();
+      // Only when the blur actually MOVED the number (TWO-25333 — see the
+      // picker's select handler for why the number and the country have to be
+      // written together). This is a BLUR, not a change: tabbing through an
+      // untouched `#company_id` fires it too, and re-pinning there would
+      // launder a stale pair into a consistent-looking one. The number would
+      // still be the previous country's company while `country_prefix` was
+      // rewritten to the country the form has since moved to, and
+      // `clearCompanyIfCountryStale` could never fire on it again. Pinning only
+      // a number the buyer actually entered keeps the witness tied to a
+      // capture rather than to a keystroke that passed through.
+      const numberMoved = typed !== Twoinc.getInstance().customerCompany.organization_number;
+      Twoinc.getInstance().customerCompany.organization_number = typed;
+      if (numberMoved) {
+        Twoinc.getInstance().customerCompany.country_prefix =
+          twoincSelectWooHelper.currentCountry();
+      }
     } else if (inputName === "billing_company_display") {
       Twoinc.getInstance().customerCompany.company_name = $input.val();
     }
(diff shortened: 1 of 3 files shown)


dgjlindsay:
## Adversarial self-review — five rounds, converged

Two independent reviewers per round (correctness lens, test-vacuity lens), each working from a freshly fetched ref. Findings by round, with severity:

| Round | Findings | Outcome |
|---|---|---|
| 1 | 1 HIGH, 4 MED, 3 LOW | re-sync branch built a two-moment pair via `getCompanyData()` (name from a sessionStorage snapshot, not the DOM) — an empty name left the method quietly unusable for a legitimately restored company; the manual blur pin laundered stale pairs; three vacuous tests |
| 2 | 1 HIGH, 3 MED, 4 LOW | round 1's fix was **fail-open** — a diverged number alone is a keystroke, not a restore; type/whitespace hazards in every comparison; unused return contract |
| 3 | 1 HIGH, 3 LOW | a documented "equivalent mutant" claim was **wrong** — the recorded-number normaliser is load-bearing, reachable via `setCompany()` writing a numeric id out of parsed JSON |
| 4 | 1 MED, 1 LOW | the two DOM-side normalisers were untested on reasoning that applied identically to the record side; survivor count described the harness, not the code |
| 5 | 1 LOW | re-synced record's stored values unpinned (they are posted verbatim in `buyer.company`); inventory still incomplete |

**Not an oscillating loop.** Rounds 3, 4 and 5 each returned one progressively less severe finding — a real laundering path, then a reachable mirror of it, then a docblock overstatement — and rounds 4 and 5 required no logic change at all. Both reviewers independently confirmed each round's findings were in code the previous round introduced but had *not* examined, rather than breakage caused by fixing something.

Three defects found here were introduced by this PR's own earlier rounds. Two were fail-open paths that a green suite would have shipped.

### Verification

- **299/299** unit tests across 8 suites (`npm run test:js`); `country-switch.test.js` carries 57, of which 30 are new.
- **Mutation-checked.** Every mutation aimed at this change is either caught by the test named in it, or listed in `tests/js/README.md` with its reason — six such, each documented in the code beside it (unreachable-from-the-only-caller guards, equivalent-by-construction reads, and one deliberate raw store that would otherwise change what the plugin posts on the order intent). The two directions the ticket demands are caught by disjoint test sets.
- **Not browser-verified.** The Chrome bridge is disconnected — checked, reports no connected extension instance — so this is unit and mutation evidence only.
- CI green apart from `e2e-tests`, which fails `admin login failed: 404` against the staging shop. Pre-existing and identical on every branch, including both PRs merged to `staging` earlier today; not caused by this change and not fixed here.

### Three things deliberately not fixed, each recorded in code

1. **`getCompanyData()` un-pins the witness** during `initialize()`'s deferred seed and for three seconds after `clearSelectedCompany`. Two-axis (country *and* name). Closing it changes that function's contract for its other callers.
2. **`setTimeout(this.enableCompanySearch, 800)`** in `initialize()` passes an unbound `this` in a strict class body. Pre-existing, fires on every real page load with company search on, and is what makes those timers unobservable to the suite.
3. **`clearSelectedCompany` re-attaches selectWoo unconditionally**, so a clear in manual-entry mode resurrects the dismissed picker. Pre-existing on the `change` path; pinned here as a characterisation assertion, labelled as such. Fixing it means changing a shared destructive function on that path too — the kind of stacking this ticket was split out of TWO-24867 to avoid.

Each wants its own ticket.

---
<sub>Review and implementation by Claude.</sub>

PR #434, opened 2026-08-03
fix(checkout): run the deferred company-search pass against the instance (TWO-25337)

dgjlindsay:
## What

`initialize()` retries `enableCompanySearch()` on an 800ms timer, to catch a billing fragment that WooCommerce had not rendered yet when `initialize()` ran. It passed the method by reference — `setTimeout(this.enableCompanySearch, 800)` — so the retry ran with the wrong receiver.

Fixed by wrapping the call, reusing the `self` that `initialize()` already declares. A `function` wrapper rather than an arrow because the file contains zero arrow functions.

## The ticket's premise was wrong, and that is the interesting part

TWO-25337 described this as a TypeError firing on every real checkout page load with company search on — a live production error. **It is not.** Measured in real Chromium before writing the fix:

```
receiver:     "window"
threw:        false
instanceProp: null
windowProp:   "ATTACHED"
console/pageerrors: []
```

`setTimeout` invokes a bare method reference with the **global** as its receiver: the timer steps supply `window` as the callback's this-value, and a strict-mode class body does not change that, because the global is only substituted for a null/undefined receiver in sloppy mode and here a receiver was passed.

The TypeError is **Sinon's**. Jest's fake timers invoke the callback with `null` where a browser passes the global, so the harness manufactures a crash no browser produces. `tests/js/README.md` had written that up as production behaviour; this PR corrects it.

The ticket has been downgraded off High accordingly.

## What the defect actually was

A misdirected write, not a crash:

- the retry set `billingCompanySelect` on `window` instead of on the singleton;
- **the widget still attached** — nothing outside `enableCompanySearch` reads that property — so the late-render retry has been working in production all along;
- what remained was a live selectWoo object leaked onto the global, plus an instance property still holding whatever the synchronous pass wrote.

That last value is **not** `null`, which is worth stating because the first draft of this PR got it wrong: `.selectWoo()` on an empty jQuery set returns that same empty set, so in the late-render case the property held a truthy widget-shaped object wrapping no element — worse than `null` for any caller testing it for presence. `null` survives only with company search off, where the early return sits above the assignment.

Behavioural delta is therefore narrow by construction: every widget-wiring call in `enableCompanySearch` either used `self` only for that one property or went through the DOM / `Twoinc.getInstance()`, so the only thing that moves is which object holds the reference.

## Tests

New `tests/js/company-search-deferred-init.test.js`, 5 tests.

| Test | Pins |
|---|---|
| stores its widget on the instance | the fix itself — discards what the synchronous pass stored, so only the retry can satisfy it |
| attaches a picker the first pass could not find | the effect the 800ms delay exists for, via a late-rendered billing fragment; also pins the truthy-empty-set value above |
| renders no second widget on the ordinary path | no duplicate container, with a witness that the retry actually ran |
| CHARACTERISATION: select handler bound twice | a real pre-existing defect (TWO-25338), recorded rather than endorsed |
| early-returns with company search off | the `enable_company_search` guard sits above the first receiver use, so those shops were and stay unaffected |

Two traps are handled explicitly, both found by review rather than by me:

**No `not.toThrow()` assertions.** All of that assertion's discriminating power here would be Sinon's, so it would go quietly vacuous the moment the timer implementation stopped manufacturing the TypeError. The tests assert the retry's effect on the instance and the DOM instead. Relatedly, the harness **cannot** reproduce the browser's `this`-binding at all — the docblock and README now say so plainly rather than letting a green run be mistaken for browser evidence.

**Witnesses against idempotence.** Most of what the retry does is idempotent: re-attaching to a field that already has a picker leaves the DOM as it was, so an end-state assertion passes whether the timer fired or not. Deleting the `setTimeout` outright left two tests green in an earlier revision, so each such test now carries a call witness.

That witness spies `Twoinc.prototype.enableCompanySearch` **before** `initialize()`, not the instance after it — which matters more than it sounds. An instance spy installed afterwards works against the wrapper shape used here (a late property lookup) but would **reject** `setTimeout(this.enableCompanySearch.bind(this), 800)`, an equally correct fix that resolves the method when the timer is scheduled. Every binding shape reaches the prototype, so the prototype spy is invariant to all of them; the cost is that the synchronous pass is counted too, hence "1 call before advancing, 2 after".

**Mutation-verified, three mutations, tree confirmed changed each time:**

| Mutation | Expected | Result |
|---|---|---|
| delete the retry entirely | all fail | 5 fail |
| revert the receiver (`setTimeout(this.enableCompanySearch, 800)`) | fail those that can see it | 4 fail |
| `setTimeout(this.enableCompanySearch.bind(this), 800)` — a *correct* alternative | all **pass** | 5 pass |

The third is the one that stops these tests from encoding my implementation choice as the only right answer. The company-search-off test passing under the receiver revert is also by design: its early return sits above the first use of the receiver, so those shops were genuinely never affected, and the test exists to say they stay that way.

**Full JS suite green: 304 tests, 9 suites.** Style gate (prettier 3.1.0) clean; `tests/js/README.md` was prettier-clean on base, so its diff stays scoped to the corrected passage.

## Out of scope, and filed: TWO-25338

Review found a real pre-existing defect that the container-count assertion was quietly reassuring about. The plugin binds its `select2:select` / `select2:open` handlers unnamespaced with no preceding `.off()`, and selectWoo's `destroy()` unbinds with `.off(".select2")` — which cannot match them — so the retry leaves both handlers bound twice on every ordinary page load.

Cost priced per handler, after review corrected me twice on it. `select2:select` is the only copy that costs anything meaningful: `renderCompanySummary()` and `togglePaySubtitleDesc()` double, `addressLookup()` doubles when address lookup is enabled (two registry round trips for one pick), the DOM and `customerCompany` writes are idempotent, and the duplicate `getApproval()` is dropped outright at the `orderIntentCheck.interval` guard — the `pendingCheck` flag it sets cannot buy a later round either, because every site that nulls the interval clears the flag in the same block.

The duplicated `select2:open` costs almost nothing, which is the opposite of what an earlier revision of this PR claimed. It looks like it should start two racing 300ms focus pollers, but the call site passes `waitToFocus("billing_company_display", null, null)` and that helper guards its defaults with `isNaN` — and `isNaN(null)` is false, so `attemptsLeft` becomes `null * 8`, i.e. 0, and the interval clears on its first tick. Two single-shot focus nudges, each already a no-op when the input is focused.

Unchanged by this PR: the retry ran on a live widget before it too. Fixing it means changing how every handler in `enableCompanySearch` is bound, which is wider than this ticket, so it is filed as TWO-25338 with the characterisation test as its regression gate.

## Verification status, honestly

**No live checkout load.** The Chrome bridge was down for this session, so the browser evidence is headless Chromium driven through the repo's own Playwright install, not a real WooCommerce checkout. That settles the `this`-binding question it was used for; the rest is the jsdom suite against the real jQuery and the real selectWoo widget.

**`e2e-tests` is red, and not from this diff.** The failure is `admin login failed: 404` in `two-api.ts` during `triggerFulfilBatch` — an auth/routing failure in the e2e harness, on a spec (order flow: place → fulfil → refund) that this JS-only change cannot reach. Every one of the last 25 `e2e-tests` runs on this repo is `failure` or `cancelled`, across unrelated PRs, so it is a systemic/environmental breakage on the shared base rather than something this branch introduced. Flagging rather than fixing: it needs whoever owns that endpoint, and this PR should not be the only copy of a fix for it. All other 17 checks pass.

Closes TWO-25337.

<sub>by Claude</sub>

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -3612,9 +3612,92 @@ class Twoinc {
       twoincSelectWooHelper.waitToFocus("billing_country");
     });
 
-    // Enable company search
+    // Enable company search, then again on a delay to catch a billing fragment
+    // that WooCommerce had not rendered yet when initialize() ran. This retry is
+    // the only one this code owns: `updated_checkout` re-syncs plenty of other
+    // state but does not re-attach the picker itself. The other callers of
+    // enableCompanySearch are the manual-entry exit and the sole-trader mode
+    // switch — so a checkout whose deferred pass misses can still be recovered
+    // by going through one of those, but nothing here is aiming for that, and it
+    // is not a path to rely on. Treat this timer as the one that has to work.
+    //
+    // Wrapped rather than passed by reference (TWO-25337). `setTimeout` invokes
+    // a bare method reference with the GLOBAL as its receiver: the timer steps
+    // supply `window` as the callback's this-value, and a strict-mode class body
+    // does not change that, because the global is only substituted for a
+    // null/undefined receiver in sloppy mode and here a receiver was passed. So
+    // `this` inside enableCompanySearch was `window`, and the deferred pass
+    // wrote its `billingCompanySelect` onto `window` rather than onto this
+    // instance.
+    //
+    // That did NOT throw and did not break the retry: nothing outside
+    // enableCompanySearch reads `billingCompanySelect`, so the widget still
+    // attached, and every other lookup in there goes through the DOM or
+    // `Twoinc.getInstance()`. What it left behind was a live selectWoo object on
+    // `window`, plus an instance property still holding whatever the
+    // synchronous pass wrote. Note what that value is, because it is NOT null:
+    // `.selectWoo()` on an empty jQuery set returns that same empty set. So in
+    // the late-render case this timer exists for, the property held a truthy
+    // empty set — a widget-shaped object wrapping no element, which is worse
+    // than null for anyone testing it for presence. (`null` survives only when
+    // company search is off, where the early return sits above the assignment.)
+    // The first reader of `this.billingCompanySelect` after a deferred
+    // re-attach would get that stale value, which is the trap
+    // `clearSelectedCompany` already avoids by looking the widget up from the
+    // DOM instead.
+    //
+    // Verified in real Chromium rather than reasoned about: the receiver is
+    // `window`, the assignment succeeds, and the console stays clean. Under
+    // Jest's fake timers the same call throws instead, because Sinon invokes
+    // the callback with `null` rather than the global — so the suite sees a
+    // TypeError that no browser ever produces. Do not restate that TypeError as
+    // production behaviour; it is a test-harness artefact.
+    //
+    // NOT fixed here, and not caused here: the deferred re-run leaves the
+    // picker's own `select2:select` / `select2:open` handlers DUPLICATED. Those
+    // are bound unnamespaced with no preceding `.off()`, and selectWoo's re-init
+    // destroys the previous instance with `.off(".select2")`, which cannot match
+    // them — so after the retry a single pick runs the whole select body twice.
+    //
+    // What that costs, per handler, neither overstated nor waved away. The
+    // `select2:select` copy is the one that costs anything.
+    //
+    // From the duplicated `select2:select`: `renderCompanySummary()` and
+    // `togglePaySubtitleDesc()` genuinely run twice, and `addressLookup()` does
+    // too when address lookup is enabled. The DOM and `customerCompany` writes
+    // are idempotent — same data both times. `getApproval()` costs nothing: the
+    // second entry finds `orderIntentCheck.interval` already set, flags
+    // `pendingCheck` and returns, and that flag cannot buy a later extra round
+    // either, because `pendingCheck` is only ever set inside that same guard and
+    // every site that nulls `interval` clears it in the same block — so the 3s
+    // poller only sees it set while the interval is still running, and the call
+    // it makes re-enters the guard and returns. The duplicate is simply dropped.
+    //
+    // From the duplicated `select2:open`: very little, and specifically NOT a
+    // pair of racing focus pollers. `addSelectWooFocusFixHandler` is idempotent
+    // (it guards on its own `two-focused-handler` attribute), and while
+    // `waitToFocus` has no dedupe, the arguments this site passes it —
+    // `("billing_company_display", null, null)` — defeat its own defaults: it
+    // guards them with `isNaN`, and `isNaN(null)` is false, so `hitsRequired`
+    // stays null and `attemptsLeft` becomes `null * 8`, i.e. 0. The interval
+    // clears itself on its first tick. Duplicating it therefore costs two
+    // single-shot focus nudges, each already a no-op when the input is focused.
+    //
+    // One stale figure to distrust while reading around this:
+    // `focusStillWithinCompanySearch`'s docblock says the poll can nudge "up to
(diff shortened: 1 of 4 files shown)


dgjlindsay:
## Adversarial self-review: 7 rounds, clean on round 7

Findings per round: **8 → 7 → 6 → 2 → 1 → 0.** Round 7 re-verified the substrate claims independently rather than only checking round 6's fix, so the clean round is a real one and not just "the last fix broke nothing".

Worth recording what the loop actually caught, because almost none of it was the code:

**The ticket's premise was wrong, and the first review round did not catch that — measurement did.** TWO-25337 asserted a live production TypeError on every checkout load. Real Chromium says the receiver is `window`, the assignment succeeds, and the console is clean. The crash only exists under Jest, because Sinon's fake timers invoke the callback with `null` where a browser passes the global. The one-line fix stayed the same; its justification and its priority both changed.

**Three factual errors in my own prose, each found by a later round:**

1. Claimed the instance property was left `null` in the late-render case. It is a truthy *empty jQuery set* — `.selectWoo()` on an empty set returns that set — which is worse than `null` for a caller testing presence.
2. Claimed the duplicate `getApproval()` cost "one redundant approval round". It costs nothing: `pendingCheck` is only ever set inside the `interval` guard, and every site that nulls the interval clears the flag in the same block.
3. Claimed the duplicated `select2:open` starts two racing 300ms focus pollers, and ranked it the worse half. It is single-shot: the call site passes explicit `null`s, `waitToFocus` guards its defaults with `isNaN`, and `isNaN(null)` is false, so `attemptsLeft` becomes `null * 8` = 0 and the interval clears on its first tick. This one contradicted a comment four lines above it, and it propagated into three places before being caught.

**Two vacuity findings, which were the most valuable part.** An early revision had tests whose assertions the *synchronous* pass already satisfied — deleting the `setTimeout` from `initialize()` entirely left them green. Worse, the first witness I added spied the instance after `initialize()`, which would have **rejected `this.enableCompanySearch.bind(this)`** — an equally correct fix, and the one the ticket itself suggested. The witness now spies the prototype before `initialize()`, so it is invariant to binding shape, and that is verified by a third mutation asserting `.bind(this)` **passes**.

Mutation matrix, tree confirmed changed each time:

| Mutation | Expected | Result |
|---|---|---|
| delete the retry | all fail | 5 fail |
| revert the receiver | fail those that can see it | 4 fail |
| `.bind(this)` (correct alternative) | all pass | 5 pass |

The company-search-off test passing under the receiver revert is by design — its early return sits above the first use of the receiver, so those shops were genuinely never affected.

**Filed rather than fixed:** TWO-25338, for the unnamespaced `select2:select` / `select2:open` handlers that survive selectWoo's `.off(".select2")` destroy and end up bound twice on every page load. Pre-existing and unchanged here — the retry ran on a live widget before this fix too. A characterisation test pins both handler counts at 2 so it cannot stay invisible, and flipping those to 1 is that ticket's regression gate.

**Docs corrected in the same PR:** `tests/js/README.md` had recorded the Sinon TypeError as a production error on every page load, and `tests/js/jest.config.js` claimed the suite used no jest mocks.

### Merging with `e2e-tests` red — deliberate

17 of 18 checks pass. `e2e-tests` fails on `admin login failed: 404` in `two-api.ts` during `triggerFulfilBatch`, in the order-flow spec (place → fulfil → refund) — a path a JS-only company-search change cannot reach. **Every one of the last 25 `e2e-tests` runs on this repo is `failure` or `cancelled`**, across unrelated PRs, so it is a systemic breakage on the shared base rather than anything this branch introduced. It is not a required check. Flagging rather than fixing, because it needs whoever owns that endpoint and this PR should not be the only copy of that fix.

<sub>by Claude</sub>

PR #436, opened 2026-08-03
feat(company-search): one control, location setting picks address area vs payment tile (TWO-25326)

dgjlindsay:
## Summary

Implements Doug's 2026-08-03 design ruling on TWO-25326 §7.1-§7.4 for WooCommerce only.

- **§7.1 — one control, admin setting picks LOCATION.** Adds a `company_search_location` admin setting (Address area / Payment tile). The existing company-search control (fields, selectWoo widget, keyboard handling — everything from §1-§4) is left untouched when the setting is "Address area" (the default, zero behaviour change). When set to "Payment tile", `twoincDomHelper.syncCompanySearchTileLocation()` relocates the SAME control (moved with `appendTo()`, never cloned) into a new `.twoinc-company-search-tile-slot` server-rendered in the payment tile.
- **§7.2/§7.3 — tile is text-only, wording embeds name/number.** Removes the standalone `.twoinc-company-tile-label` div (PR #431). The captured company now lives directly inside the intent-message sentences:
  - Approved: `This order by <name> (<number>) is likely to be accepted by <product>`
  - Declined: `<product> is not available for this order by <name> (<number>)`
  Both use the existing brand-template/token substitution mechanism (`data-company-template` + `{company}`), extended to the declined ("not available") box, which — per TWO-25224 — is never gated on the brand's notice-suppression switch.
- **§7.4 — brand-overlay wording respected.** No change forces vanilla wording over a brand's own copy: a brand overriding `intent_approved_notice` keeps its own template and just needs the token somewhere in it to get the company substitution. A new, symmetrical `intent_declined_notice` override key is added for the declined box.

## Test plan

- [x] `npm run test:js` — 306/306 passing, including a rewritten `tests/js/company-summary.test.js` covering the new intent-message substitution (both approved and declined boxes) and confirming the removed tile label stays removed.
- [x] `php tests/unit/run.php` — 169/169 passing, including updated PHP-side markup assertions.
- [x] Mutation-verified: hand-mutated the errSelector guard, the name/number formatter, and the PHP declined-template default — all three caught by the suite.
- [x] Real-browser verification (local Docker WordPress/WooCommerce against a mock Two API, since this environment has no route to a live Two backend): confirmed via the real enqueued `twoinc.js` against real WooCommerce checkout markup —
  - `payment_tile` setting: the real company-search fields (dropdown, hidden org-number field, number label) physically relocate into the tile slot, selectWoo widget intact.
  - `address_area` setting (default): control stays exactly where it always was, tile slot stays hidden.
  - Approved notice renders `This order by ACME Widgets Ltd (12345678) is likely to be accepted by Two`.
  - Declined notice renders `Two is not available for this order by ACME Widgets Ltd (12345678)`.
  - No-number (manual entry) case renders the bare name, no dangling `()`.
  - Confirmed old `.twoinc-company-tile-label` div is gone.

## Known limitation / follow-up

Live verification of the relocation surviving a genuine WooCommerce `update_checkout` AJAX re-render was not completed — this local Docker environment's mock Two API isn't faithful enough to keep that AJAX cycle from corrupting the billing-fields fragment (a pre-existing artifact of the mock, reproducible even on vanilla checkout fields unrelated to this change). The relocation re-applies on every `toggleBusinessFields()` call, which is already wired to the same `updated_checkout` event other re-anchoring logic in this file relies on (e.g. `getCompanySummaryNode()`), so it should hold up the same way — but this specific path wants a real staging re-test.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -48,6 +48,20 @@
 }
 
 /*
+ * TWO-25326 §7.3 (ruling 2026-08-03): the approved and declined intent
+ * messages now carry the captured company (`<name> (<number>)`) directly in
+ * their sentence, where the standalone `.twoinc-company-tile-label` used to.
+ * That label carried `overflow-wrap: anywhere` for exactly this reason —
+ * registry names in DE/NL/NO routinely contain a single unbroken token wider
+ * than the tile column — and these two boxes need the same protection now
+ * that they are the ones holding the name.
+ */
+.twoinc-pay-box.twoinc-intent-approved,
+.twoinc-pay-box.twoinc-err-payment-default {
+  overflow-wrap: anywhere;
+}
+
+/*
  * Visually hidden but readable by assistive technology. Defined here rather
  * than leaning on the theme's .screen-reader-text, which not every theme
  * ships — a missing rule there would render the text visibly.
@@ -160,8 +174,9 @@
  * The captured NAME no longer renders here. §7 rules out any company name or
  * number text in the address area beyond the company-name field itself and
  * this number label — the name was a second copy of what the control directly
- * above already shows. It moved to the payment tile
- * (`.twoinc-company-tile-label`).
+ * above already shows. It no longer renders as a standalone label anywhere
+ * (TWO-25326 §7.2/§7.3, ruling 2026-08-03): it lives inside the payment
+ * tile's intent-message sentence instead.
  */
 .twoinc-company-summary {
   padding-bottom: 15px;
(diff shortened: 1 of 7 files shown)


dgjlindsay:
## Adversarial review — round 1

Dispatched Leia (maintainability), Han (SRE/failure modes), Yoda (WooCommerce/jQuery idioms), Vader (edge cases) against the initial diff.

**Convergent BLOCKER (all four independently):** relocating the company-search fields directly into `.twoinc-company-search-tile-slot` left them as descendants of `.woocommerce-checkout-payment` — the exact fragment WooCommerce's `update_order_review` AJAX replaces wholesale (`replaceWith`) on a shipping-method change, coupon apply, or quantity change, not only a payment-method/country switch. A live field sitting inside that fragment at the moment of a swap was destroyed outright, with nothing to resurrect it. This matched what I'd independently observed in local live-browser testing before the review even ran.

**Fix (pushed in a follow-up commit):** `detachCompanySearchTileWrapperToSafety()`, bound to WooCommerce's own present-tense `update_checkout` trigger (fired synchronously, before the async AJAX call), moves the wrapper into a holding pen living as a direct child of `<form name="checkout">` — stable across every fragment replace — before the destructive swap can happen. `syncCompanySearchTileLocation()` (bound to the past-tense `updated_checkout`, and now called directly from `onUpdatedCheckout` rather than relying solely on `toggleBusinessFields`, per a Han finding that the latter isn't invoked on every checkout-update cycle) moves it back into the fresh slot afterwards.

Also fixed from the same round:
- Vader — `{company}` token substitution now uses a function replacer, not a string replacer, so a company name containing `$&`, `$$`, etc. can't mangle the substituted sentence.
- Yoda — missing `/* translators */` comments on the two new re-orderable `%1$s`/`%2$s` templates.
- Leia — idempotency guards on every field move (matching the existing `getCompanySummaryNode()` pattern), so a re-render that changed nothing doesn't churn a live selectWoo widget.
- Leia/Yoda — added `tests/js/company-search-tile-location.test.js`, closing the test-coverage gap both flagged (the relocation itself had zero coverage before).

Live-verified against a local Docker WordPress/WooCommerce checkout with a mock Two API: the company field and its relocated position both survive a real WooCommerce `update_checkout` AJAX cycle (tested via a country change and a plain retrigger) — confirming the fix, not just the unit test's simulation of it.

Not merging yet — running a focused round 2 against the fix itself before merge.

---
*Posted by Claude.*

dgjlindsay:
## Adversarial review — round 2 (clean)

Focused round 2 against the round-1 fix itself (Han + Vader).

**Result: no BLOCKER, no BUG.** Several suspected issues in the review brief were traced and refuted:
- `getCompanySearchTileHoldingPen()`'s `(($form.length && $form) || jQuery("body"))` fallback — traced precisely, correct (a jQuery `.length` of `0` short-circuits to falsy as intended).
- Detach-to-safety's idempotency guard — correct, no double-detach on repeated `update_checkout` firings.
- The "slot absent when re-syncing" case — wrapper stays correctly parked in the holding pen, no null-deref, no orphaned node.
- The function-replacer fix for `{company}` substitution — correct against empty string, and against a company name that itself contains the literal substring `{company}` (no re-scan/double-mangle risk).

Real findings, all OBSERVATION/NIT, none blocking:
- `stopImmediatePropagation()` on a third-party `update_checkout` handler could theoretically suppress the detach — not fixable cheaply (would need a MutationObserver belt-and-suspenders); noted as a hedge on the doc comment's certainty rather than a code change.
- WooCommerce Blocks checkout is unaddressed (this feature doesn't apply there at all — no `update_checkout`/`updated_checkout` events exist on that path) — flagged as unconfirmed-excluded, not a gap in this fix.
- Two test-coverage nits (detach-to-safety's own idempotency wasn't spy-asserted the same way sync's was; the "detach while slot absent" outcome had no test) — both added in the follow-up commit.

Both commits pushed; final full-suite run: 317 JS tests, 169 PHP tests, all passing. No further findings expected from another round on this diff — moving to merge once CI is green.

---
*Posted by Claude.*

PR #437, opened 2026-08-04
fix(company-search): drive tile location from enable_company_search, not a new setting (TWO-25326)

dgjlindsay:
## Behavior change for merchants

**Merchants who currently have "Enable company name search and auto-complete" (now "Enable Company Search In Address Entry") UNCHECKED will see a change**: company search used to be fully off in that state. It is now relocated into the Two payment tile instead of disappearing — the address-form fallback (plain, unenhanced company name field) stays exactly as it was.

## What changed

Follow-up to #436 (TWO-25326 §7.1), correcting Doug's design call: no standalone `company_search_location` setting. Instead:

- Deleted the `company_search_location` admin field, `WC_Twoinc::get_company_search_location()`, and its usage in `twoinc.js` / checkout bootstrap.
- The tile-vs-address-area decision is now derived from the pre-existing `enable_company_search` checkbox (`WC_Twoinc_Checkout::derive_company_search_location()`), preserving the `enable_company_name` back-compat fallback untouched.
- Rebadged the checkbox: **"Enable Company Search In Address Entry"** — "When enabled, the buyer may search for their company within the address entry section of the checkout. Otherwise, company search will be visible within the payment method." (matches the wording landing in magento-plugin / prestashop-plugin for cross-platform parity).
- **Bug found and fixed during verification**: the tile-relocation logic was sweeping WooCommerce's own native `billing_company` field into the payment tile alongside the search widget, so unchecking the box removed the buyer's only way to type a company name at all. Excluded the native field from the move set — Hyvä PR #95 parity (degrade to a plain field, never remove it).

## Verification

- PHP unit suite: all green (`docker run --rm -v "$PWD":/app -w /app php:8.2-cli php tests/unit/run.php`), including two new tests exercising `derive_company_search_location()` directly by reflection (mutation-verified: flipping the ternary fails the test).
- JS suite: all 318 tests green (`npm run test:js`), including a new test asserting `#billing_company_field` is never relocated (mutation-verified against the pre-fix selector).
- phpcs: 0 errors. phpstan: 0 errors.
- Local Docker (isolated stack, separate container names/ports from any existing environment): live-rendered the checkout page with the setting off — confirmed `#billing_company_field` renders in the address form at its normal position, not swept into the tile; confirmed the admin settings page shows the exact new label/description text and the old field is gone. Re-checked the default (setting on) path for no regression.

### DIFF
--- a/assets/css/twoinc.css
+++ b/assets/css/twoinc.css
@@ -606,12 +606,16 @@
 
 /*
  * The company-search control's location in the payment tile (TWO-25326 §7.1,
- * ruling 2026-08-03), when the `company_search_location` admin setting is
- * 'payment_tile'. Holds the SAME control moved out of the address area by
- * twoincDomHelper.syncCompanySearchTileLocation() — the fields themselves
- * (`#billing_company_field` etc.) already carry all their own styling from
- * the address form, so this wrapper only needs enough of its own spacing to
- * sit cleanly between the sole-trader toggle and the intent message.
+ * ruling 2026-08-03; correction 2026-08-04 folded the standalone
+ * `company_search_location` setting into the `enable_company_search`
+ * checkbox — unchecked now means this slot, not "off"). Holds
+ * `#billing_company_display_field` when twoincDomHelper.syncCompanySearchTileLocation()
+ * moves it out of the address area — NOT `#billing_company_field` or
+ * `#company_id_field`, which deliberately stay put (see that function's own
+ * doc comment). The field carries all its own styling from the address
+ * form, so this wrapper only needs enough of its own spacing to sit cleanly
+ * between the sole-trader toggle and the intent message, when it holds
+ * anything at all.
  *
  * The standalone `.twoinc-company-tile-label` text label this replaced is
  * gone outright (§7.2/§7.3): the company now lives inside the intent-message
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -1713,14 +1713,13 @@ let twoincDomHelper = {
    * See WC_Twoinc_Checkout::prepare_twoinc_object() and
    * WC_Twoinc::get_enable_company_search()'s doc comment for the source.
    *
-   * This is the same control both ways — the real `#billing_company_display_field`
-   * and `#company_id_field` rows (plus the read-only number label,
-   * `getCompanySummaryNode()`), MOVED with `appendTo()`, never cloned and
-   * never a second implementation. Whatever JS already targets those ids/
-   * classes (selectWoo init, keyboard handling, tab-order fixes,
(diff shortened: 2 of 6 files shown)


dgjlindsay:
## Adversarial review summary (self-review lane)

**Round 1** — Leia, Han, Yoda, Vader
- Han (BLOCKER): sole-trader mode's restore path could show a stale "search again" button and let `exitManualCompanyEntry()` silently flip `enable_company_search` back to `"yes"` whenever the checkbox merely read `"no"`, not only after genuine manual entry — silently overriding the merchant's setting mid-checkout. Fixed: gated on `manual_company_entry_active` too.
- Vader (BLOCKER): excluding `#billing_company_field` from the tile-move set left `#company_id_field` moving alone into an unlabelled, required tile box with no search behind it. Fixed: excluded `#company_id_field` too — both plain fields now stay together in the address form.
- Yoda: `company_search_location` (PR #436) was never added to `drop_removed_settings()`. Fixed, with a new unit test mirroring the existing `enable_sole_trader` case.
- Self-caught while fixing Vader's finding: excluding `#company_id_field` broke `getCompanySummaryNode()`'s idempotency (it anchors against `#company_id_field`'s current location), creating an infinite address-form/tile re-anchor loop. Removed the summary-node tile-relocation entirely.
- Leia (nits): stray double-blank line, one redundant `get_enable_company_search()` call — both fixed.

**Round 2** — Han, Vader (targeted re-check)
- Vader: clean, round-1 fixes hold, no new findings.
- Han (BUG): the round-1 fixes left the tile-move loop a no-op in the only reachable state (checkbox off), but the slot was still unconditionally unhidden — every checkbox-off merchant got a bare, unexplained ~24px gap in the payment tile. Fixed: only unhide when the wrapper actually gained a child. New JS test reproduces the real server-side DOM shape and is mutation-verified.

**Round 3** — Han (convergence check): clean, round-2 fix holds, no new findings (one out-of-scope, pre-existing, harmless duplicate-property observation, not touched by this diff).

**Verification**: PHP unit suite (319 assertions incl. new reflection-based tests) and JS suite (319 tests) green; phpcs 0 errors; phpstan 0 errors; live-verified three times in isolated local Docker stacks (separate from any other running environment) — checkout render with the setting on/off, admin settings page wording, and the orphaned-option cleanup on upgrade.

**Open item flagged to Doug (not a blocker for this PR)**: the admin field's mandated description ("company search will be visible within the payment method") isn't literally true in WooCommerce today — the control degrades to a plain fallback in the address area with an empty tile, since making search functionally live inside the tile is materially bigger scope (touches the search-init gate, server-side field registration, and required-field semantics) than this correction. Scoped this PR to the field-safety fix; left for a follow-up decision.

— by Claude

PR #322, opened 2026-06-10
TWO-24747/feat: Local dev harness targets and env plumbing

dgjlindsay:
## What

D1 of the plugin parity plan (TWO-24739 → TWO-24747). **Ticket premise correction**: the compose stack (wordpress + wpcli provisioner + mariadb, product seeding, plugin activation, config JSON) already existed — this PR adds the missing ergonomics, not the environment.

**Stacked on #321** (needs the brand layer's `TWO_BRAND_CODE` handling).

## How

- `.env.example` → copy to `.env` (compose's native env file — the ticket's `.env.local` spelling diverged from the repo's existing `.env` convention; code wins): `TWO_API_BASE_URL`, `TWO_API_KEY`, `TWO_BRAND_CODE`, WooCommerce knobs.
- Compose passes `TWO_BRAND_CODE` into the WordPress container (PHP `getenv` → `WC_Twoinc_Brand`) and parameterises the provisioning vars with the previous hardcoded values as defaults.
- `dev/configure`: idempotent wp-cli script writing `TWO_API_KEY`/`TWO_API_BASE_URL` onto the gateway settings; runs at first provision and via `make configure`.
- `docker/wpcli.sh` now derives the settings option key from the gateway id (was hardcoded — a brand-override run would have seeded the wrong option row).
- Makefile: `install` / `run` / `configure` / `logs` / `logs-wpcli` / `stop` / `clean` / `test-unit` / `test` / `format`. `make test-unit` runs the TWO-24744 unit harness in a `php:8.2` container — the ticket's "establish PHPUnit baseline" is superseded by that suite (deliberately dependency-free; CI already runs it on 7.4+8.2).

## Test plan

- [x] `docker compose config` valid; `bash -n` on both scripts
- [x] `make test-unit` green (19/19)
- [ ] Fresh `make clean && make install` provision on a dev box, checkout loads, `make configure` round-trips an api-key change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #438, opened 2026-08-04
fix(checkout): declined intent notice must never be brand-overridable

dgjlindsay:
## Summary
Per the 2026-08-04 TWO-25326 ruling, the buyer-facing "order intent NOT approved" checkout tile notice must render identical platform-default copy for every brand — only the approved/success notice is meant to be brand-overridable.

`get_intent_declined_notice_template()` previously read `WC_Twoinc_Brand::get('intent_declined_notice')`, mirroring the approved notice's copy-override mechanism. No shipped brand file used the key, but the seam was live and reachable end-to-end (same defect shape found and fixed in `two-inc/magento-plugin` PR #322).

- `class/WC_Twoinc.php`: `get_intent_declined_notice_template()` always renders the literal default copy; no brand-key read.
- `brands/two.php`: documents the deliberate omission so it doesn't get "helpfully" re-added.
- New fixture `tests/unit/fixtures/decidedoverridebrand.php` + test `testDeclinedNoticeIgnoresABrandOverrideKey` lock in that a brand file declaring the key is silently ignored.

The approved-notice override (`intent_approved_notice`) is unchanged — the companion ABN-overlay fix (`two-inc/woocommerce-abn-plugin`) turns that ON with ABN copy instead of leaving it suppressed.

## Test plan
- [x] `make test-unit` — full suite green, including the new test.
- [x] `php -l` on every touched file.

### DIFF
--- a/brands/two.php
+++ b/brands/two.php
@@ -109,4 +109,10 @@ return [
     // the brand product_name and %2$s the buyer's company name.
     // WC_Twoinc::get_intent_approved_notice is the only reader.
     'intent_approved_notice' => null,
+    // Deliberately no 'intent_declined_notice' key here (removed
+    // 2026-08-04 ruling, TWO-25326): the "order intent NOT approved"
+    // notice is never brand-overridable. WC_Twoinc::
+    // get_intent_declined_notice_template() no longer reads any brand
+    // key for this — do not add one back, and do not add one to an
+    // overlay's own brand file either; it would be silently ignored.
 ];
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -919,22 +919,22 @@ if (!class_exists('WC_Twoinc')) {
          *
          * Same token mechanism as get_intent_approved_notice(): %1$s is the
          * brand product name, %2$s is the company token twoinc.js substitutes
-         * with the same "<name> (<number>)" (or bare name) string. A brand
-         * may override the template via 'intent_declined_notice', mirroring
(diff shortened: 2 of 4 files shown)


dgjlindsay:
## Adversarial self-review — round 1 (clean)

**Scope:** removes the dead `intent_declined_notice` brand-override seam in `get_intent_declined_notice_template()` and `brands/two.php`, matching the ruling that the "not available" notice must never be brand-overridable; adds a locking test (`testDeclinedNoticeIgnoresABrandOverrideKey`) with a dedicated fixture that declares the key to prove it's ignored.

**Checks performed:**
- Grepped the full branch for `intent_declined_notice` — no remaining reads outside comments/the new fixture/test. The seam is fully removed, not just unused.
- Confirmed `twoinc_brand_file` is an established, real filter hook used consistently by other tests in `run.php` — the new test isn't inventing a new seam.
- Ran the full unit suite locally (`php tests/unit/run.php`) against the PR branch — all tests pass, including the new one.
- **Mutation check:** reverted `get_intent_declined_notice_template()` to the old brand-override-reading behavior and reran — `testDeclinedNoticeIgnoresABrandOverrideKey` failed as expected, confirming the new test actually locks the ruling rather than being vacuous.
- Diffed against `origin/staging` for brand/partner name leaks (public repo) — none found; fixture uses a generic placeholder name (`decidedoverridebrand`).

No findings. Round is clean.

---
*by Claude*

PR #439, opened 2026-08-04
fix(company-search): make payment-tile placement functional (TWO-25326)

dgjlindsay:
## What

Fixes TWO-25326 item §7.1/#64.1: unchecking "Enable Company Search In Address Entry" is supposed to relocate the ONE company-search control into the payment tile, never turn it off. `WC_Twoinc_Checkout::update_company_fields()` still gated the control's registration on that checkbox, so unchecking it left the tile's relocation JS (`syncCompanySearchTileLocation()`) with nothing to move — the tile rendered empty and buyers saw only the plain, unenhanced fallback fields with no working search anywhere on the page.

## Change

- `update_company_fields()` now registers `billing_company_display` unconditionally. Only its mount location (address area vs payment tile) is still driven by the checkbox, via the existing `company_search_location` derivation.
- Decouples `window.twoinc.enable_company_search`'s two prior meanings:
  - the admin's raw checkbox preference — now read via `company_search_location` wherever that's what's meant (the three `enable_company_search_for_others` gates, which the setting's own description ties to address-area placement)
  - the JS runtime "is the search widget the active input method right now" flag — now always `"yes"` at page load, since the control is never fully off; still toggled to `"no"` by manual-entry and sole-trader mode exactly as before.
- Widens `syncCompanySearchTileLocation()`'s unhide guard from "wrapper has a child" to "wrapper has a *visible* child" — manual-entry/sole-trader mode now leaves a real (but `hidden`-classed) field in the wrapper instead of no field existing at all, and the old guard would have unhidden an empty-looking tile around it.
- Corrected several doc comments across `WC_Twoinc_Checkout.php`, `twoinc.js` and the Jest suite that had documented the old gate as permanent, intended behaviour.

No block-checkout code path exists in this plugin to update — `WC_Twoinc_Checkout` hooks the classic `woocommerce_checkout_fields` filter only; there's no WooCommerce Blocks integration in this repo.

## Testing

- `php tests/unit/run.php` — full suite green, plus a new test asserting `billing_company_display` is registered regardless of `get_enable_company_search()`'s return value (`yes`/`no`/`null`/`''`).
- `npm run test:js` — 320/320 green. Extended `company-search-tile-location.test.js` with:
  - a test driving the real `Twoinc.initialize()` wiring and asserting the relocated control is a live, initialized selectWoo widget (`$(...).data("select2")` truthy) while sitting inside the tile slot — not just a moved DOM node
  - an updated hidden-field guard test reflecting the field's new always-present, sometimes-`hidden` DOM shape

Confirmed exactly one company-search control definition in the codebase (`billing_company_display` / `#billing_company_display_field`) — no second parallel implementation was added for the tile.

Not merging — leaving for review per standing process.

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -748,7 +748,7 @@ let twoincSelectWooHelper = {
       if (this.tabIndex < 0) return;
       if (jQuery(this).closest(".select2-container--open, .select2-dropdown").length) return;
       if (twoincSelectWooHelper.isHiddenForTabbing(this)) return;
-      if (!((anchor.compareDocumentPosition(this) & 4) /* DOCUMENT_POSITION_FOLLOWING */)) return;
+      if (!(anchor.compareDocumentPosition(this) & 4 /* DOCUMENT_POSITION_FOLLOWING */)) return;
       found.push(this);
     });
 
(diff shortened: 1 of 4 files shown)


dgjlindsay:

Ran a fresh adversarial round on top of the prior rounds already reflected in the code comments (Vader/Han/Doug's rulings, 2026-08-04):

- **State-machine trace** of `enable_company_search` (runtime "is search the active input method" flag), `company_search_location` (admin placement preference), and `enable_company_search_for_others` (address-area-only gate) across every read/write site in `twoinc.js` and `WC_Twoinc_Checkout.php` — no remaining collision between the old and new meanings.
- **WooCommerce checkout-semantics trace** — `billing_company_display` is now always registered with `required => false`, so always-on registration can't fail WC's required-field validation regardless of placement/visibility. The dynamic-selectWoo-vs-static-options concern is pre-existing (predates #439) and unaffected by this change.
- **Hidden-field slot-visibility trace** — confirmed `toggleBusinessFields()` only ever uses `.hidden`/`.removeClass("hidden")` (never `.hide()`/`.show()`), so the widened `$wrapper.children(":not(.hidden)").length` guard in `syncCompanySearchTileLocation()` correctly catches the manual-entry/sole-trader hidden-field case, and call ordering relative to `toggleBusinessFields()` is intact.
- **Security/DOM lens** — no new dynamic selector construction or unescaped output; nothing to flag.

Also found and fixed a real, unrelated build failure: the diff had incidentally dropped a paren prettier 3.1.0 (the pinned pre-commit version) requires around the `compareDocumentPosition` bitmask check in `assets/js/twoinc.js` — pre-commit was red on this PR. Restored it in a follow-up commit; verified against `prettier@3.1.0` locally to match CI exactly, then reran the full local Jest (320/320) and PHP unit suites (`tests/unit/run.php`, all pass) before pushing.

Verdict: clean round — no findings from this pass, and the round whose only "finding" was the prettier fix does not carry forward any new logic risk (pure formatting, verified against pinned tool version + full test suite).

—Claude

PR #440, opened 2026-08-04
refactor(company-search): encapsulate WC search/dropdown/manual-entry into TwoCompanySearch class (TWO-25326)

dgjlindsay:
## What

Rebuilds the WooCommerce company-search JS to follow PrestaShop's
`TwoCompanySearch` discipline: a single class owns the entire
search/dropdown/manual-entry/select2 lifecycle, constructed once with a
config object, rather than a flat helper object plus scattered functions
elsewhere in the file.

- `twoincSelectWooHelper` is now the one `TwoCompanySearch` instance
  (`new TwoCompanySearch({ companyFieldSelector: "#billing_company_display" })`),
  constructed exactly once at module load — the single call site.
- Company-search-specific `twoincDomHelper` functions (payment-tile
  relocation, manual entry, the read-only company summary, dropdown
  open/focus, `getCompanyName`) moved onto the class. Checkout-wide concerns
  (business-field visibility, sole-trader mode, checkout-input
  snapshotting) stay in `twoincDomHelper`/`twoincSoleTrader` and call into
  the class, the same way `Twoinc#enableCompanySearch` now does.
- Added `TwoCompanySearch#attach()` — the one place selectWoo is
  initialised against the configured field and the `select2:select`/
  `select2:open` handlers are wired up. `Twoinc#enableCompanySearch` is now
  a thin gate + delegation to it.

This is an architecture-only refactor — no behaviour change. Both
address-area and payment-tile placement (PR #439) continue to work exactly
as before.

## Verification

- `grep -c "^class TwoCompanySearch"` → 1; `grep -c "new TwoCompanySearch("` → 1
- `npm run test:js` — 319/319 passing (test files updated to call the moved
  methods on the class instance instead of `twoincDomHelper`; no assertions
  or fixtures changed)
- PHP unit suite (`make test-unit`) — all passing, unaffected (no PHP
  behaviour changed; one doc-comment reference updated)
- `prettier --check` clean

## Scope

Do not merge — stopping here per request for review.

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -75,7 +75,42 @@ let twoincUtilHelper = {
   }
 };
 
-let twoincSelectWooHelper = {
+/**
+ * Company-search widget: search/dropdown/manual-entry/select2 lifecycle,
+ * encapsulated (TWO-25326 architecture rebuild). Mirrors PrestaShop's
+ * TwoCompanySearch class — a single class owns the ENTIRE search, dropdown,
+ * manual-entry and select2 lifecycle, configured via a constructor options
+ * object, with exactly one construction site (below, `twoincSelectWooHelper`).
+ * Same class, same dropdown/query-field/manual-entry behaviour, never a
+ * second implementation.
+ *
+ * Company-search-adjacent but checkout-wide concerns — toggling which
+ * fields are visible for the selected account type, the intent-message
+ * text, sole-trader mode — stay in `twoincDomHelper` / `twoincSoleTrader`
+ * below and call into this class's public methods, the same way
+ * `Twoinc#enableCompanySearch()` does.
+ */
+class TwoCompanySearch {
+  /**
+   * @param {Object} [options]
+   * @param {string} [options.companyFieldSelector] CSS selector of the
+   *   company-search <select> to attach to. Defaults to
+   *   `#billing_company_display`, the id WooCommerce always renders it
+   *   under on this plugin's checkout.
+   */
+  constructor(options) {
+    options = options || {};
+    this.companyFieldSelector = options.companyFieldSelector || "#billing_company_display";
+  }
+
+  /**
+   * CSS selector of the <select> this instance attaches selectWoo to. Set
+   * from the constructor's options above; declared here (rather than left
+   * purely as an instance property) so every other field this class owns is
+   * visible in one place.
+   */
+  companyFieldSelector;
+
   /**
    * Hard ceiling on a single company-search request, ms (TWO-25232). Before
    * this there was no client timeout at all, so a request that never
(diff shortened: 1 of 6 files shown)


dgjlindsay:
## Adversarial review — round 1 (clean)

Verified independently against the current diff and the PR-branch file, not just the agent's self-report.

**PR #439 visibility guard preserved correctly.** The `$wrapper.children(":not(.hidden)").length` tile-visibility guard from #439 lives inside `syncCompanySearchTileLocation()`, now a `TwoCompanySearch` method (`assets/js/twoinc.js:1649` on the PR branch). Confirmed byte-for-byte unchanged logic, just moved with the rest of the method — no rebase-conflict damage.

**Single-construction discipline holds.** Grepped for a second `new TwoCompanySearch(...)` — none. Class is defined once (`class TwoCompanySearch {`), constructed once at module load (`let twoincSelectWooHelper = new TwoCompanySearch({...})`), matching PrestaShop's discipline as claimed.

**No stale cross-references.** Checked every method that moved off `twoincDomHelper` onto the class (`genSelectWooParams`, `attach`, `bindManualEntryAffordance`, `syncCompanySearchTileLocation`, `getCompanyName`, `clearSelectedCompany`, `getCompanySummaryNode`, `renderCompanySummary`, tab-trap helpers, etc.) for any leftover `twoincDomHelper.<method>` call site — zero found. `togglePaySubtitleDesc` and other genuinely checkout-wide concerns correctly stayed on `twoincDomHelper`.

**`attach()` double-init risk — investigated, not a regression.** `enableCompanySearch()` (and therefore `attach()`) is called from 4 sites: `initialize()`, its 800ms retry, `exitManualCompanyEntry()`, and the sole-trader mode switch. Each call re-runs `$field.selectWoo(...)` and rebinds `select2:select`/`select2:open` via `widget.on(...)`. Diffed this against the pre-refactor inline code (`git show` on the parent commit) — it is the exact same call pattern, just relocated from inline `Twoinc#enableCompanySearch` into `TwoCompanySearch#attach`. Not something this PR introduced or worsened; flagging only as a pre-existing architectural trust in selectWoo's own re-init teardown, unchanged by this refactor.

**Address-area / tile-mode parity.** Both `syncCompanySearchTileLocation()` (relocation) and `detachCompanySearchTileWrapperToSafety()` (AJAX-destruction guard, TWO-25326 §7.1 hardening) are intact and correctly still keyed off `twoincSelectWooHelper.*`, matching the wiring in `Twoinc` (`$body.on("update_checkout", twoincSelectWooHelper.detachCompanySearchTileWrapperToSafety)`).

**Verification run:**
- `node -c assets/js/twoinc.js` — syntax OK.
- `npm run test:js` on the PR-branch worktree (post-rebase HEAD `e4365ad`): **10 suites / 320 tests passed.**
- Test-file diff is pure `ctx.dom.*` → `ctx.helper.*` renames tracking the moved methods — no assertion or fixture changes, matches the PR description.
- The one PHP change (`class/WC_Twoinc_Checkout.php`) is a doc-comment update only, no behavior change.

No findings. Clean round.

*by Claude*

PR #441, opened 2026-08-04
fix(company-search): keep summary span with relocated control, keep native field on checkbox-off (TWO-25326)

dgjlindsay:
## Summary

Two live-verified bugs in the company-search payment-tile relocation (built on #440's `TwoCompanySearch` class):

1. **Orphaned summary span.** `getCompanySummaryNode()` anchored the read-only org-number summary (`#twoinc_company_summary` / `.twoinc-company-summary-id`) against `#company_id_field`, which is deliberately never relocated into the tile. Moving the search control into the tile therefore left the summary behind as a stray element in the address section. Fix: anchor against `#billing_company_display_field` instead — the field that actually relocates — so the summary follows it into the tile or stays in the address area, whichever is current.

2. **Native field removed on checkbox-off.** `toggleBusinessFields()` decided whether to show the search field or WooCommerce's own native `#billing_company_field` using the runtime `window.twoinc.enable_company_search` flag. That flag is now hardcoded to `"yes"` by `prepare_twoinc_object()` so the relocated widget stays functional regardless of the admin checkbox — but it meant the native-field branch could never fire, so unchecking "Enable Company Search In Address Entry" removed the native field from the address area entirely instead of just moving the search control. Fix: show the native field alongside the relocated search control whenever `company_search_location` is `'payment_tile'` — the two are independent concerns (search control placement vs. WooCommerce's own field existing).

## Test plan

- [x] `npm run test:js` — 322/322 passing, including two new regression tests in `tests/js/company-search-tile-location.test.js` (one per bug)
- [x] Both new tests verified to fail against the pre-fix code and pass against the fix (non-vacuous)
- [x] `php tests/unit/run.php` — all passing (no PHP changes in this PR)
- [x] `npx prettier --check` clean

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -817,7 +817,7 @@ class TwoCompanySearch {
       if (this.tabIndex < 0) return;
       if (jQuery(this).closest(".select2-container--open, .select2-dropdown").length) return;
       if (twoincSelectWooHelper.isHiddenForTabbing(this)) return;
-      if (!(anchor.compareDocumentPosition(this) & 4 /* DOCUMENT_POSITION_FOLLOWING */)) return;
+      if (!((anchor.compareDocumentPosition(this) & 4) /* DOCUMENT_POSITION_FOLLOWING */)) return;
       found.push(this);
     });
 
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -817,7 +817,7 @@ class TwoCompanySearch {
       if (this.tabIndex < 0) return;
       if (jQuery(this).closest(".select2-container--open, .select2-dropdown").length) return;
       if (twoincSelectWooHelper.isHiddenForTabbing(this)) return;
-      if (!((anchor.compareDocumentPosition(this) & 4) /* DOCUMENT_POSITION_FOLLOWING */)) return;
+      if (!(anchor.compareDocumentPosition(this) & 4 /* DOCUMENT_POSITION_FOLLOWING */)) return;
       found.push(this);
     });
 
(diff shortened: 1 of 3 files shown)


dgjlindsay:
## Adversarial self-review — clean

Reviewed the diff against the two bug reports it fixes, plus one independent second-pass review.

**Bug 1 (orphaned summary span):** confirmed `getCompanySummaryNode()`'s old anchor (`#company_id_field`) never relocates, so pre-fix the read-only summary genuinely stayed behind in the address area once the search control moved to the tile. New anchor chain (`#billing_company_display_field` → `#company_id_field` → `#billing_company_field`) tracks the field that actually moves.

**Bug 2 (native field removed on checkbox-off):** confirmed `#billing_company_field` is never part of the relocation set — `syncCompanySearchTileLocation()` only ever moves `#billing_company_display_field`; `#billing_company_field`/`#company_id_field` are explicitly excluded and stay in the address form. So the fix (`toggleBusinessFields()` un-hiding `#billing_company_field` when `company_search_location === "payment_tile"`) leaves the native field exactly where WooCommerce puts it, with the search control additionally live in the tile — matches the exact requirement (native field untouched/address-area, search control tile-only), no duplicate-field confusion within the tile.

**Regression tests:** traced both new tests in `company-search-tile-location.test.js` against pre-fix behavior — both genuinely fail without the fix (not vacuous): the summary-anchor test would see `closest(".twoinc-company-search-tile-slot")` == 0 pre-fix, and the native-field test would see `#billing_company_field` still carrying `.hidden` pre-fix. `company-summary.test.js`'s updated anchor assertions match the new anchor order.

**Also fixed:** a stray formatting change to an unrelated line (`compareDocumentPosition` bitwise-AND parenthesization) had slipped into the original commit and tripped CI's pre-commit/prettier check — CI pins `prettier@3.1.0`, which disagrees with the locally-installed 3.9.6 on this expression's parens. Reverted that line back to match `staging` in a follow-up commit; unrelated to either bug fix.

No blocking findings from either review pass. One NIT (comment verbosity, matches existing file convention) not worth holding on.

— by Claude

PR #442, opened 2026-08-04
fix(company-search): re-decide tile visibility on payment-method switch (TWO-25326)

dgjlindsay:
## Bug 1 — search control not visible in payment tile

Root cause: with "Enable Company Search In Address Entry" unchecked (`company_search_location: payment_tile`), the search control never appeared in the payment tile at all when the buyer switched TO the Two gateway from a different default gateway — the ordinary case, since WooCommerce checks the first available gateway by default.

`toggleBusinessFields()` is the function that decides whether `#billing_company_display_field` is shown. `onUpdatedCheckout()` only relocates whatever is already there via `syncCompanySearchTileLocation()`; the only rebind of `toggleBusinessFields()` to a `payment_method` change lived *inside* `onUpdatedCheckout()` itself, with no matching `.off()` — bound too late to catch the first switch of a session (nothing forces `updated_checkout` to have fired before a buyer picks a payment method), and it accumulated a duplicate on every cycle besides. `syncCompanySearchTileLocation()`'s own "unhide only if a visible child moved in" guard then correctly kept the tile slot hidden around a still-hidden field — the exact symptom reported.

Fix: a single namespaced, delegated `payment_method` change listener bound once in `initialize()`, calling `toggleBusinessFields()` directly. The late/duplicating rebind is removed from `onUpdatedCheckout()`.

New regression test: `tests/js/company-search-payment-method-switch.test.js` drives the real bootstrap IIFE (the existing harness deliberately leaves it un-exercised) and fails against the pre-fix code — verified locally by reverting the JS change and re-running.

## Requirement 2 — remove "Enable company name search for other payment options"

Removed entirely per Doug's ruling: a second, independent toggle never made sense — it now just follows the main "Enable Company Search In Address Entry" checkbox (`company_search_location === 'address_area'`). Removed:
- the admin field registration
- the JS reads (`window.twoinc.enable_company_search_for_others`), replaced with a read of `company_search_location`
- the admin.js visibility wiring
- the README sample config entry

`drop_removed_settings()` purges the option key from upgraded installs (same mechanism as `enable_sole_trader`/`company_search_location`), covered by a new PHP test.

## Testing

- `npm run test:js`: 11 suites / 323 tests, all pass
- `php tests/unit/run.php`: all pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,6 @@ cat > docker/config/staging.json <<EOF
   "show_abt_link": "yes",
   "section_auto_complete_settings": "",
   "enable_company_search": "yes",
-  "enable_company_search_for_others": "yes",
   "enable_address_lookup": "yes"
 }
 EOF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -43,10 +43,6 @@ jQuery(function ($) {
     function (e) {
       toggleChildrenFields(
         $(this),
-        $("#woocommerce_" + twoinc_admin.gateway_id + "_enable_company_search_for_others")
-      );
-      toggleChildrenFields(
-        $(this),
         $("#woocommerce_" + twoinc_admin.gateway_id + "_enable_address_lookup")
       );
     }
@@ -72,10 +68,6 @@ jQuery(function ($) {
 
   jQuery("h3.wc-settings-sub-title").next().hide();
 
-  toggleChildrenFields(
-    $("#woocommerce_" + twoinc_admin.gateway_id + "_enable_company_search"),
-    $("#woocommerce_" + twoinc_admin.gateway_id + "_enable_company_search_for_others")
-  );
   toggleChildrenFields(
     $("#woocommerce_" + twoinc_admin.gateway_id + "_enable_company_search"),
     $("#woocommerce_" + twoinc_admin.gateway_id + "_enable_address_lookup")
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -2573,10 +2573,18 @@ let twoincDomHelper = {
         }
       }
     } else {
+      // Whether company search is available for OTHER payment methods is no
+      // longer a setting of its own (TWO-25326, Doug's ruling: a separate
+      // `enable_company_search_for_others` toggle was one control too many —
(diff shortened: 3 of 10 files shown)


dgjlindsay:
**Adversarial self-review (self-review lane, `reviewDecision` null)**

1. **Double-binding risk** — OK. `initialize()` guards its whole body with an `isInitialized` flag set only at the end, so the new namespaced delegated `change` listener (`change.twoincPaymentMethod` on `document.body`) is registered exactly once per page load even though `initialize()` is called from ~9 call sites. The `.off()` before `.on()` is defensive, not load-bearing.

2. **Other code relying on the removed un-`.off()`'d binding** — OK. The removed binding inside `onUpdatedCheckout()` was a single, self-contained `change` handler calling `toggleBusinessFields()`; nothing else in that span depended on it, and the surrounding calls (`updateElements()`, `rearrangeDescription()`) are untouched.

3. **Semantic equivalence of `company_search_location === 'address_area'` vs the old setting** — real behavior change, flagged for visibility rather than a bug: `company_search_location` derives solely from the main `enable_company_search` checkbox. The removed `enable_company_search_for_others` was a genuinely independent stored option. A merchant who had explicitly set the main checkbox to `yes` and set "for others" to `no` will, after this upgrade (and the `drop_removed_settings()` purge), silently get search shown everywhere instead of suppressed on non-Two gateways. This matches the PR's stated intent to retire the setting, so treating as accepted rather than blocking, but noting it here since it's a genuine behavior delta for any merchant who had diverged the two settings.

4. **Upgrade-purge test** — OK, non-vacuous. `testEnableCompanySearchForOthersSettingDroppedFromUpgradedInstalls` seeds the stale option (plus an unrelated option to keep) before invoking `drop_removed_settings()`, then asserts only the stale option is gone — would fail against a no-op purge.

5. **Repo-wide leftover references to `enable_company_search_for_others`** — OK, zero live reads left. Remaining hits are comments, the removal-list literal in `drop_removed_settings()`, and the test that verifies removal. No hits in README, POT/PO, or CHANGELOG.

CI: all green except `e2e-tests` still running at review time.

Clean round — no blocking findings. Merging once e2e passes and the standard 10-minute self-review window has elapsed.

— by Claude

PR #443, opened 2026-08-04
fix(company-search): sentence-case address-entry search label (TWO-25326)

dgjlindsay:
Sentence-case the "Enable Company Search In Address Entry" admin label to match house style.

## What changed
- `class/WC_Twoinc.php`: label -> "Enable company search in address entry" (no trailing period)
- `tests/unit/run.php`: doc-comment references to the old string updated to match

## Convention check
PrestaShop's equivalent label was sentence-cased in PR #138. Checked other admin labels on the same WooCommerce settings page — most verb-phrase labels already use sentence case with no trailing period (e.g. "Skip user validation at order confirmation", "Pre-approve buyer during checkout", "Clear settings on deactivation of plug-in", "Display input tooltips"). No label on the page carries a trailing period, so the period was dropped from Doug's literal wording to stay consistent with the actual codebase convention.

No translation files (.pot/.po) reference this string yet — it hadn't been extracted for translation. No test asserts the exact string as a value; only doc-comment prose quoted it, which was updated for accuracy.

## Test plan
- [x] `make test-unit` — all pass
- [x] `npm run test:js` — 323 passed

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -2537,7 +2537,7 @@ let twoincDomHelper = {
         // WooCommerce's OWN native company field is a completely separate
         // concern from where OUR search control lives (bug found by Doug
         // 2026-08-04, live-verified against the checkbox-off/payment_tile
-        // state): unchecking "Enable Company Search In Address Entry" moves
+        // state): unchecking "Enable company search in address entry" moves
         // the search control into the payment tile, but must never take
         // WooCommerce's stock field away from the address area — the two
         // coexist, search in the tile, native field where WC always puts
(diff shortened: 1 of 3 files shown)


dgjlindsay:
**Adversarial review — TWO-25326 sentence-case rename**

Checked:
- Repo-wide grep for the old string `"Enable Company Search In Address Entry"` — found and fixed 3 leftover comment references in `assets/js/twoinc.js` that the original diff missed (it updated comments in `class/WC_Twoinc.php` and `tests/unit/run.php` but not this file). Not a functional break (comments only, no exact-string lookups), but fixed for consistency since matching the label's casing across comments was the point of the PR. Now 0 remaining references to the old casing anywhere in the repo.
- Confirmed no `.pot`/`.po` file uses either the old or new string as a msgid/translation key — this is not a translated label, so no i18n catalog impact.
- No other exact-string lookups (JS, tests, admin templates) depend on the literal label text.

Round 2 (post-fix): clean, nothing new found.

By Claude.

PR #444, opened 2026-08-05
fix: clear stale merchant ID/short name when API key verification fails

dgjlindsay:
## Summary

The admin settings page kept showing the previously-fetched merchant ID and
short name even after the stored API key failed re-verification. `verify_api_key()`
only persists `merchant_id` / `merchant_short_name` on a 200 response and
leaves the cached option values untouched on any failure, and the page-load
JS re-verification (`admin.js`) only toggled the red/green status icon —
never touching the merchant info panel — so a broken key left the settings
page looking healthy.

- Added a hidden-by-default "API key could not be verified" notice next to
  the existing merchant-info / signup-prompt blocks in
  `generate_api_key_with_verification_html()`.
- `admin.js` now hides the merchant-info panel and shows that notice whenever
  the verify AJAX call reports failure or errors at the transport level, on
  both page load and live key-typing verification.

## Test plan

- [x] `npx jest --config tests/js/jest.config.js` — full JS suite green (326 tests), including new `tests/js/api-key-verification.test.js` covering: verify failure, transport error, and successful verification.
- [x] `php tests/unit/run.php` — full PHP suite green, unaffected by this change.
- [x] Manual read-through of `generate_api_key_with_verification_html()` render output and `admin.js` DOM wiring to confirm the notice element IDs match.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -107,6 +107,18 @@ jQuery(function ($) {
     $("#twoinc-merchant-short-name").text(shortName ? " · " + shortName : "");
     $("#twoinc-merchant-info").show();
     $("#twoinc-signup-prompt").hide();
+    $("#twoinc-merchant-invalid-notice").hide();
+  }
+
+  // A failed verification must not leave the previously fetched Merchant ID
+  // on screen — that reads as "the integration is fine" when it isn't. Swap
+  // it for an explicit invalid-key notice instead (this call site is the fix
+  // for the settings page silently keeping stale merchant info on a broken
+  // key).
+  function showMerchantInfoInvalid() {
+    $("#twoinc-merchant-info").hide();
+    $("#twoinc-signup-prompt").hide();
+    $("#twoinc-merchant-invalid-notice").show();
   }
 
   function verifyApiKey(apiKey) {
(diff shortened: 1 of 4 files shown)


dgjlindsay:
## Self-review

Diff is small (4 files, ~130 LOC incl. tests), so this was a single-pass adversarial self-review (correctness, edge cases, framework-convention checks) rather than the full multi-persona dispatch, per the adversarial-review skill's own guidance on proportionality for this size.

**Round 1 — found 1 real bug, fixed:**
- The new "API key could not be verified" notice hardcoded the literal string "Two" instead of `WC_Twoinc_Brand::get('product_name')`, breaking the white-label overlay convention every other user-facing string in this file follows. Fixed in `f47d283`.

**Round 1 — checked, no issue:**
- No-key / fresh-install path: `merchant_id` renders empty, the new notice div defaults to `display:none`, no regression.
- Successful re-verification after a prior invalid state correctly re-shows merchant info and hides the notice (covered by test 3).
- XSS: the new string is static, translated, no interpolated user data — no escaping needed.
- Test harness change (`apiKey`/`stubAjax` in `admin-harness.js`) is additive and gated behind `options.apiKey === undefined`, so it doesn't change behavior for any existing suite that doesn't opt in — confirmed by running the full `tests/js` suite (326/326 green) both before and after.

**Round 2 — re-reviewed the diff after the brand-name fix: nothing new found.** Clean round.

Also ran `php tests/unit/run.php` (full suite green, unaffected by this change) and `npx jest --config tests/js/jest.config.js` (326/326 green, including the new `api-key-verification.test.js`).

— by Claude

PR #445, opened 2026-08-05
fix(api-key): differentiate verification failures and hide checkout on failure

dgjlindsay:
## Summary

Today's incident showed every non-200 response from `/v1/merchant/verify_api_key` — an actual 401/403, a Two 5xx, or a network/routing failure that never got a response at all — was treated identically as "API key is invalid" everywhere in the plugin. That made a routing failure indistinguishable from a merchant typo, and the Two payment method + address-block company-search control kept rendering at checkout regardless of whether the key could actually be verified.

- `WC_Twoinc::categorize_verification_result()` buckets a `verify_api_key()` result into `ok` / `invalid_key` (401/403) / `service_error` (5xx) / `unreachable` (network/timeout) / `not_configured` / `error` (other), shared by the settings-page AJAX handler and the new cached status check. No raw response body is surfaced anywhere — only the category and HTTP status.
- `get_api_key_verification_status()` caches that categorization behind a short (5 min) transient keyed by the API key itself, so `is_available()` and the checkout script bootstrap don't fire a live HTTP call on every evaluation — mirrors the TTL-cache shape already used by `WC_Twoinc_Sole_Trader`.
- `WC_Twoinc::is_available()` now requires a verified key (on top of the existing enabled check), so the payment method is not offered at checkout when verification fails for ANY reason.
- `WC_Twoinc_Checkout::inject_cart_details()` withholds `window.twoinc` entirely on the same condition. Since the whole company-search/payment-tile bootstrap in `twoinc.js` is gated behind `if (window.twoinc)`, this is also what stops the address-block company-search control from rendering/enabling itself on a broken integration.
- `admin.js` now shows the categorized reason (not the raw response body/content) in a new `#twoinc-merchant-invalid-notice` element on the settings page, so an admin can tell "my key is wrong" from "Two's API is down" without guessing.

Scope: WooCommerce only, per plan — Magento/Hyvä/PrestaShop parity is a later round.

Note: PR #444 (open, unmerged) touched the same settings-page markup/JS area for a related but narrower fix (clearing stale merchant info on failure). This PR's `#twoinc-merchant-invalid-notice` element supersedes that one with dynamic, category-specific text — landing whichever merges first will produce a small, easy merge conflict on the other; flagging for a merge-order call rather than closing #444 myself.

## Test plan

- [x] PHP unit suite (`php tests/unit/run.php`) green, including new tests for categorization, TTL caching, `is_available()` (both failure and success), and checkout-side suppression on failure.
- [x] JS suite (`npx jest --config tests/js/jest.config.js`) green, including a new suite asserting the admin notice text differs per failure category (invalid key vs 5xx vs network) and is absent on success.
- [x] `prettier --check` clean on touched JS.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -110,15 +110,50 @@ jQuery(function ($) {
     $("#twoinc-merchant-invalid-notice").hide();
   }
 
+  // Maps the categorized failure the AJAX handler reports (see
+  // WC_Twoinc::categorize_verification_result()) to admin-facing text — no
+  // raw response body/content, just enough to tell "my key is wrong" apart
+  // from "Two's API is down" (TWO-25326 follow-up: today's incident showed
+  // every non-200 response, 5xx and network failures included, displayed
+  // identically as "API key is invalid").
+  function invalidNoticeText(status, code) {
+    switch (status) {
+      case "invalid_key":
+        return "This API key is invalid or has expired.";
+      case "service_error":
+        return (
+          "Two's API returned a service error (HTTP " +
+          code +
+          "). This is likely temporary on Two's side — try again shortly."
+        );
+      case "unreachable":
+        return "Could not reach Two's API (network or connectivity error). Try again shortly.";
+      case "not_configured":
+        return "Enter an API key above to enable Two.";
+      case "request_failed":
+        // The AJAX request to THIS SITE's admin-ajax.php failed before Two
+        // was ever contacted (jQuery's error callback fires on any
+        // transport-level failure — a WordPress-side 500, an expired
+        // nonce, a proxy/WAF block — not specifically "Two is
+        // unreachable"). Reviewed round 1: the "unreachable" wording here
+        // would have wrongly pointed an admin at Two for a WP-side problem.
+        return "Could not complete verification — try again shortly.";
+      default:
+        return code
+          ? "Two's API returned an unexpected response (HTTP " + code + ")."
+          : "This API key could not be verified.";
+    }
+  }
+
   // A failed verification must not leave the previously fetched Merchant ID
-  // on screen — that reads as "the integration is fine" when it isn't. Swap
-  // it for an explicit invalid-key notice instead (this call site is the fix
-  // for the settings page silently keeping stale merchant info on a broken
-  // key).
-  function showMerchantInfoInvalid() {
+  // on screen — that reads as "the integration is fine" when it isn't.
+  // Swap it for the categorized invalid-key notice instead (this call site
+  // is also the fix for the settings page silently keeping stale merchant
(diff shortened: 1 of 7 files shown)


dgjlindsay:
## Self-review (adversarial, 2 rounds)

**Round 1** — 4 personas (Leia/maintainability, Han/SRE, Vader/adversarial QA, Yoda/WP-WC idiom) reviewed the initial diff in parallel. Findings, all fixed in `49a0a3a`:

- **Han**: the admin settings page's own live re-verification never fed the checkout-side cache, so a merchant who fixed a broken key could see checkout stay hidden for up to 5 min after the fix. Fixed via a shared `cache_verification_result()` both paths now write through.
- **Yoda (blocker)**: `is_available()`'s cache-miss path could inherit `verify_api_key()`'s 30s admin-page timeout, blocking a customer-facing page render for up to 30s if Two is unreachable. Fixed with an explicit 5s timeout, mirroring `WC_Twoinc_FX::FETCH_TIMEOUT`.
- **Yoda**: the new transient key was a bare literal instead of going through `WC_Twoinc_Brand::prefixed_name()` like `WC_Twoinc_FX` does. Fixed.
- **Yoda**: `is_available()` returning `false` on verification failure had no log trail — the exact "couldn't tell why" gap this PR exists to fix, just moved to an unlogged surface. Fixed with a log-once-per-request line, mirroring the existing `apply_brand_availability_gate()` pattern.
- **Vader**: a truthy-but-malformed API response (no `body` key) fell through to the same `null` verify_api_key() returns for "not configured" — wrong message for a response that was actually received. Fixed with an explicit `malformed_response` branch.
- **Leia**: admin.js's ajax `error` callback (fires on any transport failure reaching `admin-ajax.php`, not specifically Two) hardcoded "Could not reach Two's API" wording. Fixed with a neutral `request_failed` category.
- **Leia (nit)**: removed `is_api_key_verified()`, a one-line wrapper with no remaining callers after inlining into `is_available()`.

**Round 2** — verified each fix against the actual current code (not the diff) and ran the real test/lint commands myself. All seven hold; no new correctness issues. Two accepted, non-blocking observations:
- The AJAX handler now also warms the cache for a typed-but-unsaved candidate key, not just the saved one — harmless (5 min TTL, keyed by that key's own hash).
- The `static $logged` once-per-process guard in the new `is_available()` logging isn't itself asserted by a test — matching the existing, identical pattern in `apply_brand_availability_gate()`, which the codebase has never tested either (adding a test for it would be order-dependent across the whole suite and break the moment tests are reordered).

Full PHP (`php tests/unit/run.php`) and JS (`npx jest --config tests/js/jest.config.js`) suites green throughout; `prettier --check` clean.

— by Claude

PR #446, opened 2026-08-05
fix(i18n): make API key verification notices translatable (TWO-25326)

dgjlindsay:
## Summary
- PR #445 added six admin-notice failure messages (invalid/expired key, 5xx service error, connection error, "enter API key" prompt, unexpected-response, plus the generic fallback) to `admin.js` as hardcoded English strings, and dropped the previously-translatable "could not be verified" string from `WC_Twoinc.php`.
- Follows the existing `days_label` pattern: localize the categorized notice text via `wp_localize_script` (`api_key_notices`), and have `invalidNoticeText()` read from it, falling back to English only if the localized data didn't arrive.
- Adds the 7 new msgids to the `.pot` and machine-translates them into `nb_NO`, `nl_NL`, `sv_SE`; recompiles the corresponding `.mo` files.

## Test plan
- [x] `php -l class/WC_Twoinc.php`
- [x] `node --check assets/js/admin.js`
- [x] `.github/scripts/check-catalogues.sh` passes (all 3 catalogues agree with their `.po` sources)


dgjlindsay:
**Self-review summary (by Claude)**

Small, mechanical i18n-only diff — reviewed inline rather than dispatching full adversarial personas (per `two-review-adversarial`'s own guidance for <50 LOC mechanical changes).

Checked:
- All 7 `api_key_notices` PHP strings byte-match their JS fallback literals in `invalidNoticeText()` — a missing localization silently degrades to the same English text rather than to a different string.
- `%s` → HTTP-status substitution (`service_error`, `unexpected_response`) verified against the only two switch branches that pass a `code`.
- `.pot`/`.po`/`.mo` all agree (`check-catalogues.sh` green) across nb_NO, nl_NL, sv_SE.
- Added translator comments for both `%s` placeholders, matching the existing convention elsewhere in this catalogue (e.g. the company-search `%d` comment), so translators know it's a status code, not free text.
- `php -l`, `node --check`, and the full CI matrix all pass.

No blockers found. Round complete — no further findings.

PR #447, opened 2026-08-05
Sentence-case surcharge tax admin labels + translations

dgjlindsay:
## What changed
- `class/WC_Twoinc.php`: "Surcharge Tax Treatment" -> "Surcharge tax treatment", "Surcharge Tax Class" -> "Surcharge tax class" (TWO-25326), matching the sentence-case convention set in PR #443.
- Grepped the repo (PHP/JS/templates, doc comments, tests) for any other Title-Case occurrence of these two strings — found none. Dropdown option values inside `get_surcharge_tax_treatment_options()` / `get_surcharge_tax_class_options()` were left untouched (out of scope).

## Translations
Neither msgid was present in `languages/*.pot`/`*.po` before this change (not yet extracted for translation), so no translation files needed updating — same finding as PR #443 for its label.

## Tests
No existing test asserts the exact label text (tests reference the field *keys*, e.g. `surcharge_tax_treatment`, not the title strings), so there was no test-text convention to mirror — same as PR #443's finding.

## Test plan
- [x] `make test-unit` — all pass

### DIFF
--- a/class/WC_Twoinc.php
+++ b/class/WC_Twoinc.php
@@ -4252,7 +4252,7 @@ if (!class_exists('WC_Twoinc')) {
                     'default'     => ''
                 ],
                 'surcharge_tax_treatment' => [
-                    'title'       => __('Surcharge Tax Treatment', 'twoinc-payment-gateway'),
+                    'title'       => __('Surcharge tax treatment', 'twoinc-payment-gateway'),
                     'desc_tip'    => __('How the surcharge line is taxed. Standard applies your store\'s default tax rules to the fee. Specific tax class taxes the fee under a WooCommerce tax class you select below — to leave the fee untaxed, create a tax class with a 0% rate and select it here.', 'twoinc-payment-gateway'),
                     // Visible (non-tooltip) description: carries the loud
                     // never-taxed error when the stored treatment is one this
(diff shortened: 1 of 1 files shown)


dgjlindsay:
## Adversarial self-review — round 1 (clean)

**Correctness of casing changes**
- Confirmed both target strings changed exactly as specified: `Surcharge Tax Treatment` -> `Surcharge tax treatment`, `Surcharge Tax Class` -> `Surcharge tax class`. No trailing punctuation added, matching PR #443's convention.

**Scope check — out-of-scope option-value strings**
- Verified `get_surcharge_tax_treatment_options()` and `get_surcharge_tax_class_options()` are untouched (`git diff origin/staging...HEAD` shows zero hits inside either function).

**Missed Title-Case strings**
- Grepped the whole repo (PHP/JS/templates/tests/doc comments) for both target strings — no other occurrence found.
- Noted but deliberately left alone (out of scope per the brief, which named only these two labels): sibling fields on the same settings block, `Surcharge Calculation Basis` and `Surcharge Line Description` (`class/WC_Twoinc.php`), are also Title Case. Flagging for a possible follow-up ticket rather than fixing here, to keep this diff to exactly what was asked.

**Translations**
- Neither msgid existed in `languages/*.pot`/`*.po` before this change (confirmed via `grep -c "Surcharge Tax" languages/*`, 0 hits in all four files pre-change) — same situation PR #443 hit for its label. No translation files needed updating; nothing to get wrong on terminology since nothing was extracted yet.

**Test coverage**
- No test in `tests/unit/run.php` asserts the exact label text (existing tests reference the field keys `surcharge_tax_treatment` / `surcharge_tax_class`, not the title strings) — same finding as PR #443, no test-text convention exists to mirror.
- `make test-unit` passes in full (all suites, no regressions).

**CI**
- All 18 checks green: unit-tests 7.4–8.4, phpcs, phpstan, pre-commit, e2e-tests, Jest, boot/upgrade smoke, gettext catalogues.

No findings from this round. Recommending merge.

[Claude Sonnet 5] by Claude

PR #449, opened 2026-08-05
fix(i18n): resolve the product name in API key notices from the brand (TWO-25326)

dgjlindsay:
## Problem

PR #445 added six categorized API-key-verification failure notices, and PR #446 made them translatable. Four of the seven messages write the literal product name `Two` into the copy:

| key | before |
| --- | --- |
| `service_error` | `Two's API returned a service error (HTTP %s). This is likely temporary on Two's side — try again shortly.` |
| `unreachable` | `Could not reach Two's API (network or connectivity error). Try again shortly.` |
| `not_configured` | `Enter an API key above to enable Two.` |
| `unexpected_response` | `Two's API returned an unexpected response (HTTP %s).` |

Correct for the vanilla plugin; wrong for a brand overlay, whose admin gets told to wait for — and blame — a product that overlay does not ship. PR #444's own round-1 self-review had already caught and fixed exactly this in the sibling "could not be verified" notice, by routing the name through `WC_Twoinc_Brand::get('product_name')`. This applies that established fix to the four the follow-ups added.

## Fix

| key | after (msgid) |
| --- | --- |
| `service_error` | `%1$s's API returned a service error (HTTP %2$s). This is likely temporary on %1$s's side — try again shortly.` |
| `unreachable` | `Could not reach %s's API (network or connectivity error). Try again shortly.` |
| `not_configured` | `Enter an API key above to enable %s.` |
| `unexpected_response` | `%1$s's API returned an unexpected response (HTTP %2$s).` |

- `WC_Twoinc::api_key_notice_templates()` holds the source copy with the product name as a placeholder. `get_api_key_notices()` fills it from `WC_Twoinc_Brand::get('product_name')` — the convention every other user-facing string in that file already follows.
- The HTTP status code is not known at localize time (it arrives with the AJAX response), so the two status-bearing messages pass a literal `'%s'` as their *status* argument. `sprintf()` does not rescan its own output, so the brand is substituted server-side and the `%s` survives for `admin.js` to replace with the status code — the `days_label` pattern, unchanged on the JS side.
- `admin.js`'s fallback literals (reachable only if the localisation never arrived) are now brand-neutral. That file ships unchanged to overlays, so a product name hardcoded there is the same bug one layer down.
- The four msgids changed, so the `.pot` and the `nb_NO` / `nl_NL` / `sv_SE` `.po` catalogues are re-translated against the new source strings and the `.mo` files recompiled.

## Hardening the self-review found on top of it

The notice copy now passes through `sprintf()`, and its format string is whatever catalogue is installed **at runtime** — a Loco-edited `.mo` or a translate.wordpress.org import can carry a placeholder the source never had. `WC_Twoinc::format_api_key_notice()` is the containment for that, and it is most of the diff:

- **Both PHP failure shapes.** PHP 8 throws `ArgumentCountError`/`ValueError`; PHP 7.4 (the plugin's floor, `phpstan.neon` `phpVersion: 70400`) emits a warning and returns `false`. Unhandled, the first takes down the whole gateway settings page from `admin_enqueue_scripts`, the second puts a bare `false` into the notice. The verdict is taken from the *result* (`@vsprintf` + `is_string`) with the catch kept for PHP 8. **The `unit-tests (7.4)` CI leg was genuinely red on this** — `make test-unit` only runs php:8.2, so the local gate could not see it.
- **A dropped placeholder.** `vsprintf()` accepts a format string that uses *fewer* arguments than it was given, so a translation that simply omitted the status placeholder formatted cleanly and the notice silently lost the HTTP status code. The two status-bearing categories now require it to survive.
- **A placeholder nothing will fill.** `strip_unfilled_placeholders()` drops a specifier from copy the plugin does not substitute into (the three argument-free notices, the two status-free formatted ones, and the degraded fallback) rather than printing it at an admin. It deliberately leaves an escaped `%%` and a prose percent (`100% of the time`) alone; the docblock says plainly which residue that trade accepts.
- **A bounded, diagnosable log.** These notices are rebuilt on *every* wp-admin page view (the hook has no screen check) and a bad catalogue stays bad, so the error is throttled per category through a transient, and names the category — four can break at once.
- **CI.** `msgfmt --check-format` added to `.github/scripts/check-catalogues.sh`; plain `msgfmt` compiles a `msgstr` whose placeholder set differs from its `msgid` without complaint, and this PR doubles the placeholder-bearing entries and introduces positional specifiers. A hand-rolled `msgid`-vs-`msgstr` comparison was also written to cover the entries `msgfmt` skips (it only inspects `#, php-format` entries) and then **reverted** — it false-positived on plural entries, URL-encoded text and fuzzy entries, and by that script's own header a gate that fails unrelated PRs is worse than no gate. The gap is documented in place; the runtime strip is what covers that case.
- `AGENTS.md` described the catalogue gate as only catching `.mo`/`.po` drift and prescribed recompiling the `.mo`, which is the wrong fix for the header and placeholder failures it also reports.

## Tests

- `testApiKeyNoticesUseOverlayProductNameNotTwo` — under the overlay brand fixture, all four notices name the overlay and none names the vanilla product; the `%s` the JS substitutes survives the brand interpolation.
- `testApiKeyNoticesCarryTwoProductNameAndStatusPlaceholder` — pins the rendered vanilla English, so this fix cannot silently reword the default copy.
- `testApiKeyNoticeCatalogueWithBadPlaceholdersDegradesNotFatals` — degrade rather than fatal, one log line **per broken category** naming it, throttled by a transient (asserted directly — a per-request static would satisfy a call-count assertion and still log on every page view), and a healthy category left un-throttled so its first real failure is not suppressed.
- `testApiKeyNoticesNeverShipAnUnfillablePlaceholder` — the strip covers `%s`, `%1$s`, `%d` and the `%%s`-unescaped-by-formatting case, and leaves prose percents alone.
- `testApiKeyNoticeDroppingTheStatusPlaceholderDegrades` — the residual-status gate.
- `testApiKeyNoticeCopyIsTranslatedInEveryLocale` — reads the msgids off the live source (not retyped) and checks each is present in the `.pot` and in every compiled `.mo`, plus a per-locale copy fragment. Mirrors the existing `skip_confirm_auth` precedent; catches the source being reworded without the catalogues being regenerated.
- JS: one test that the no-localisation fallback names no brand, and one that a repeated status placeholder is fully substituted. The harness gained an opt-in `apiKeyNotices` option so the default world keeps covering the fallback path.

Every PHP assertion above was verified to fail against the unfixed code, by mutation. Stated precisely, because it changes what the tests are worth: the three JS *brand-copy* tests pass against pre-fix `admin.js` too — preferring localized copy over a literal is behaviour it already had, and the brand fix itself lives in PHP. They are retitled to say so, and stand as regression guards on the interpolation contract rather than evidence for this change.

Unit suite verified green on **every PHP version CI runs** (7.4, 8.0, 8.1, 8.2, 8.3, 8.4), not just the local 8.2. `phpcs`, `phpstan`, `shellcheck`, `prettier`, Jest and the catalogue gate green locally.

Five adversarial self-review rounds; round 2 caught the red 7.4 leg, round 5 caught that its own round-4 CI addition was a false-positive machine.

Linear: TWO-25326


dgjlindsay:
## Adversarial review — round 6 (clean)

Reviewed the full diff independently against the stated goal (de-hardcode "Two" from the four API-key-verification notices added in PR #445/#446, brand-aware via `WC_Twoinc_Brand::get('product_name')`) plus the hardening rounds 1-5 already produced.

Checked:
- **Brand fix itself**: `WC_Twoinc::get_api_key_notices()` resolves `product_name` from `WC_Twoinc_Brand`, same convention as the sibling `unverified` notice PR #444 fixed. `admin.js`'s fallback literals (only reached when localisation never arrives) are brand-neutral — confirmed no residual hardcoded "Two" anywhere in the changed JS or PHP paths.
- **Placeholder-safety hardening** (`format_api_key_notice`, `strip_unfilled_placeholders`): handles both PHP 8 (`ArgumentCountError`/`ValueError`) and PHP 7.4 (`false`-returning `vsprintf`) failure shapes — confirmed `unit-tests (7.4)` CI leg is green, which is what the PR body says was the thing the local `make test-unit` (php:8.2 only) couldn't have seen. Degraded fallback is itself run through the strip, so a broken fallback catalogue can't leak a placeholder either.
- **Log throttling**: asserted via the transient directly, not just a call count — matches the PR's own stated rationale (a per-request static would pass a naive assertion while still logging every page view).
- **`.github/scripts/check-catalogues.sh`**: added `msgfmt --check-format` gate; the reverted hand-rolled msgid/msgstr comparison left no dead code behind — checked the full script, it's clean.
- **Translations**: `.pot`/`.po`/`.mo` for nb_NO/nl_NL/sv_SE all carry the reworded msgids with the placeholder-numbered forms (`%1$s`/`%2$s`) where needed; `testApiKeyNoticeCopyIsTranslatedInEveryLocale` reads msgids off live source rather than a retyped copy, so it can't drift silently.
- **Tests**: PHP suite covers the overlay-brand path, the vanilla-brand pin, catalogue-mismatch degradation (both "extra placeholder" and "dropped placeholder" shapes), and the never-ship-unfillable-placeholder guard across `%s`/`%1$s`/`%d`/escaped-`%%`. JS suite covers the same brand/no-brand contract from the admin.js side, plus a repeated-placeholder substitution case. No test asserts on implementation details it can't independently justify.
- **Leak gate**: no partner/private names introduced; "ABN" is not mentioned; the overlay fixture uses a generic "Testbrand" name rather than any real brand.
- **Scope**: everything in the diff traces back to the sprintf-based brand interpolation this fix necessarily introduces (the hardening isn't unrelated gold-plating — it's the failure modes that interpolation itself opens up).

No new findings. CI is green across all legs including `unit-tests` (7.4-8.4), `Jest`, `phpcs`, `phpstan`, `Gettext catalogues`, and `e2e-tests`.

— by Claude

PR #450, opened 2026-08-05
fix(company-number): never display TWO:-prefixed org numbers (TWO-25326)

dgjlindsay:
## What

Two requirements from TWO-25326, checked together. Only one needed code.

### Requirement 12 — never display a `TWO:`-prefixed organisation number (FIXED)

Sole-trader enrollment mints an organisation number of the form `TWO:…` for a buyer with no registry identifier of their own. It is a protocol value the backend derives the company type from, not a number the buyer's own authorities would recognise, so it must not reach any surface a human reads.

It still has to be **posted and sent to the API** — the payment method cannot authorise an order without one — so this filters the *display* only. Every test asserts both halves: suppressed on screen AND still present on the submitted field. A test that only checked the former would pass just as happily against a fix that dropped the value.

**One shared formatter, not a filter per call site**, added in `twoincUtilHelper` beside `blankToEmpty`:

| Helper | Job |
|---|---|
| `isSyntheticCompanyNumber(value)` | The predicate, for callers that branch on the fact rather than render the value. Empty is deliberately **not** synthetic — "nothing captured yet" is a different state from "a number that must not be shown", and only the latter hides a field. |
| `formatCompanyNumber(value)` | The display value; `""` for a minted one, so every caller's existing truthiness check doubles as the suppression. |
| `composeCompanyLabel(label, value)` | The `<label> (<number>)` chunk. **The brackets belong to the number**: when it resolves to nothing the label comes back bare, never `Example Ltd ()`. |
| `formatCompanyLabel(name, value)` | `composeCompanyLabel` for a plain-text name. |

The composer takes its label **untouched** because the two callers disagree about what it holds — the intent notices pass plain text bound for `.text()`, the search dropdown passes the search response's pre-highlighted HTML fragment bound for `innerHTML`. Trimming or re-encoding centrally would have to pick one contract and break the other, so plain-text callers collapse their own name via `formatCompanyLabel` and the fragment caller uses `composeCompanyLabel`.

**Four display sites routed through it:**

- **The summary label under the company field** (`renderCompanySummary`) — the block is already keyed on the number being non-empty, so it now hides itself rather than showing an empty strip.
- **The search-results dropdown rows** — while `company_id` keeps the raw value, so selecting the row still captures the company.
- **The approved and declined order-intent notices**, via `getCompanyLabelText` — both had the parens written as literal text around a value that can now come back empty. This is the `Company Name (123456789)` → `Company Name` case.
- **`#company_id` itself**, in the one configuration that shows it to the buyer: company search off, in a country whose registry supports sole traders. The field is **hidden rather than blanked**, because the value is what WooCommerce posts and what the intent is authorised against — the enrollment already supplied it, so there is nothing left to type. Its required cue goes with it; a required cue on an invisible field is how a checkout becomes unsubmittable with no visible reason why.

### Requirement 11 — spinner must be local, not a page-covering overlay (ALREADY CORRECT, no change)

The order-intent spinner is already scoped to the payment tile and was left alone.

It is a `<div class="twoinc-pay-box twoinc-loader" role="status">` injected into the payment-method box's own description (`class/WC_Twoinc.php`, `get_intent_loader_html()`), revealed by `jQuery(".twoinc-pay-box.twoinc-loader").removeClass("hidden")` in `togglePaySubtitleDesc("checking-intent")` (`assets/js/twoinc.js`). Its only rule in `assets/css/twoinc.css` is `text-align: center` — no `position: fixed`, no backdrop, no mask. The plugin calls no `blockUI`/`.block()` anywhere, and the company-search spinner is likewise a `position: absolute` element inside the dropdown's own search box.

Verified by grep across the plugin's JS, CSS and PHP for `blockUI`, `.block(`, `position: fixed`, overlay and mask patterns: no page-covering spinner exists to fix.

## Translation

**No new or changed user-facing strings.** Requirement 12 is pure filtering of values that were already on screen, and requirement 11 needed no change at all — so no `.po`/`.json` catalogue entries were added and no locale files are touched.

## Tests

New `tests/js/company-number-display.test.js`, 27 cases: the helpers directly (including the brackets-omitted-when-empty case and the empty-is-not-synthetic distinction), plus each of the four display sites end to end, each paired with an assertion that the submitted value survived.

```
Test Suites: 13 passed, 13 total
Tests:       354 passed, 354 total
```

**Mutation-checked, not just green.** Stubbing `isSyntheticCompanyNumber` to `return false` fails 10 of the new tests — at least one per display site — so they genuinely pin the filter rather than passing vacuously. That check also caught a real vacuity in the first draft: the harness checkout form carries no `payment_method` radio, so `isTwoincVisible()`/`isTwoincSelected()` read false and every company display was hidden *for that reason* instead of the one under test. The suite now appends a checked radio, which is why the `#company_id` assertions are meaningful.

## Deliberately NOT changed — needs a ruling

The **admin user-profile "Billing Company ID" input** (`class/WC_Twoinc.php`) also renders a stored `company_id`, and can hold a minted one. It is left as-is on purpose: it is an *editable* field whose value round-trips through `save_user_meta()`, which writes `$_POST['twoinc_company_id']` unconditionally. So blanking the rendered value makes the next profile save erase the stored identifier, and omitting the input makes that same save wipe it to empty.

It is a merchant-admin data-editing screen rather than a buyer-facing display, so it is outside the spirit of §12, and the safe fix needs a decision about the save path (guard on `isset`, or round-trip the real value through a hidden field). Flagged rather than guessed at.

No PHP formatter was added because no *checkout* PHP renders an organisation number value — `WC_Twoinc_Checkout.php` passes it to the browser as data and renders only the field's label.

### DIFF
--- a/assets/js/twoinc.js
+++ b/assets/js/twoinc.js
@@ -36,6 +36,105 @@ let twoincUtilHelper = {
   },
 
   /**
+   * Prefix marking an organisation number as internally minted rather than
+   * issued by a company registry (TWO-25326 §12).
+   *
+   * Sole-trader enrollment mints one of these for a buyer who has no registry
+   * identifier of their own; it carries the company-type semantics the backend
+   * derives on its side. It is a protocol value, not a number the buyer's own
+   * authorities would recognise, so showing it to them is meaningless at best
+   * and reads as a corrupted field at worst.
+   *
+   * Matched as a literal, case-sensitive prefix: it is a string the plugin's
+   * own backend mints, not something a buyer or a registry ever types, so
+   * there is no case or spacing variant to be liberal about. A real
+   * organisation number that merely CONTAINS these characters somewhere other
+   * than the start is not one of these and stays visible.
+   */
+  SYNTHETIC_NUMBER_PREFIX: "TWO:",
+
+  /**
+   * Is this organisation number internally minted, i.e. one §12 forbids
+   * showing (TWO-25326 §12)?
+   *
+   * For the callers that need to branch on the fact itself rather than just
+   * render the filtered value — hiding a whole field, say. Empty is NOT
+   * synthetic: "no number captured yet" is a different state from "a number
+   * that must not be shown", and only the latter suppresses a field the buyer
+   * would otherwise type into.
+   *
+   * @param {*} value
+   * @returns {boolean}
+   */
+  isSyntheticCompanyNumber: function (value) {
+    return twoincUtilHelper
+      .blankToEmpty(value)
+      .startsWith(twoincUtilHelper.SYNTHETIC_NUMBER_PREFIX);
+  },
+
+  /**
+   * Normalise an organisation number for DISPLAY (TWO-25326 §12).
+   *
+   * Returns `""` — never the prefixed value — for an internally minted
+   * identifier, so every caller's existing "is there a number" truthiness
+   * check doubles as the suppression: a label keyed on it hides itself, a
+   * bracketed composition drops its brackets, and no call site needs a filter
+   * of its own.
+   *
+   * DISPLAY ONLY. The value still has to be posted and sent to the API —
+   * Two's payment method cannot authorise an order without it — so the
+   * submitted `#company_id` input, the instance state and the order-intent
+   * payload all keep the raw value. Anything reading this function's result
+   * is, by construction, about to put it in front of a human.
+   *
+   * @param {*} value
+   * @returns {string} the number to show, or `""` if it must not be shown
+   */
+  formatCompanyNumber: function (value) {
+    if (twoincUtilHelper.isSyntheticCompanyNumber(value)) return "";
+    return twoincUtilHelper.blankToEmpty(value);
+  },
+
+  /**
+   * Compose the "<label> (<number>)" chunk both company displays use, with
+   * the number filtered through formatCompanyNumber (TWO-25326 §12).
+   *
+   * The brackets belong to the number, not to the composition: when the
+   * number resolves to nothing the label is returned bare, never with an
+   * empty pair of parens trailing it. That is the whole reason this is one
+   * function rather than a filter applied at each call site — two of the
+   * three sites had already written the parens as literal text around a
+   * value that can now come back empty.
+   *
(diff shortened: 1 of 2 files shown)


dgjlindsay:
## Review — TWO-25326 §12 org-number display suppression

Diffed against current `origin/staging` (post-#449 merge). Verified the fix broadly, not just the touched call sites:

- Confirmed all four display sites route through the shared `formatCompanyNumber`/`composeCompanyLabel`/`formatCompanyLabel` helpers: `renderCompanySummary` (summary label), the search-results dropdown (`item.highlight`/`item.name` composition), both order-intent notices via `getCompanyLabelText`, and `#company_id_field`'s own visibility toggle (hidden rather than blanked, correctly — it's the posted/submitted value).
- Grepped the whole repo (not just the diff) for other rendering sites: `views/woocommerce_order_pay.php` only emits a static input, `WC_Twoinc_Checkout.php`'s `fields['billing']['company_id']` is a static label ("Company ID"), `WC_Twoinc_Helper.php`'s `organization_number` references are all API-payload/meta plumbing, not display. No PHP template renders the value. Found no missed second/third call site.
- Bracket-removal logic is correct: `composeCompanyLabel` only appends `" (" + number + ")"` when `number` is truthy after filtering, so a suppressed number collapses to the bare label rather than `Name ()`. `isSyntheticCompanyNumber` correctly treats empty as NOT synthetic (distinct "not captured yet" vs "must not be shown" states) — verified via test cases.
- The admin user-profile "Billing Company ID" field is deliberately left unfiltered per the PR body's own flagged reasoning (editable field whose value round-trips through unconditional `save_user_meta()`); agree this is out of scope for §12 and correctly called out for a separate ruling rather than guessed at.
- Prefix match is case-sensitive by design — correct, since `TWO:` is a literal value the plugin's own backend mints, not user input with case variance.

**Tests:** Jest 13 suites / 365 tests passed (`npm run test:js`). PHP unit harness (`php tests/unit/run.php`) — all tests passed (only unrelated PHP 8.5 `ReflectionProperty::setAccessible()` deprecation noise). CI: all 18 checks green.

**Leak-gate:** clean — no partner/merchant names in diff, commits, branch name, or PR title/body.

No issues found. Merging.

— by Claude

